PR #3137, opened 2024-03-08
[BEAM-3896] USAM Migration

### LINKED ISSUE #3163: USAM- Migration flow
The shape of Microservices in Beamable SDK 2.0 looks very different from SDK 1.0. We must force a migration from the old format to the new format. Dotnet and CLI installation is already included in the SDK and can be taken as valid assumptions. 

We must be able to detect the presence of a all old Microservice in the project. The following rules were used to determine the existence of Microservices in 1.0,

A class that extends Microservice, and

has a [Microservice] attribute, and

has an associated .asmdef in the same folder.

On every domain reload, we should check for the existence of these files using a Reflection Cache. If any are found, we must show a popup explaining a data migration is going to occur, and that the developer should ensure they have a backup of their project. 

The following steps must be taken for each Microservice

create a hidden Microservice folder in the location the .asmdef used to be,

populate it with the required service files by using the beam project add command,

copy the entire contents of the folder into the hidden folder's service folder. However, if there are subfolders with their own .asmdef files, do not copy those folders, or their decendents. This means we need to write our own recursive file copier.

create a sign post asset for the service

automatically create .beam.csproj files for any referenced .asmdef files from the original .asmdef file

delete the original .asmdef,

delete all the original source code in the folder, including all sub folders that don't have their own .asmdef files.

A similar process must happen for all Microstorage objects.

Leinnan:
# Ticket

https://disruptorbeam.atlassian.net/browse/BEAM-3896

# Brief Description

Fixes #3163 and #3170

Replaces `beam project new` and `beam project add` with single command: `beam project new-microservice`. Also makes the flow and arguments of `beam project new-storage` more in line with those in `beam project new-microservice`


cdhanna:
I would like to see an explanation for the workflows. 
How do you create a new project?
How do you create a new .beamable folder + project at the same time?
How do you add a project to existing .beamable folder? 

We'll probably need to update some of the files in the /docs sub folder.

Leinnan:
@chrisbeamable 
> How do you create a new project?

beam project new-microservice SuperService

> How do you create a new .beamable folder + project at the same time?

beam project new-microservice SuperService

> How do you add a project to existing .beamable folder?

What do you mean by "add"? If create a new microservice and add it to existing one it would be:

beam project new-microservice SecondService --existing-solution-file Service.sln

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -95,6 +96,105 @@ public async Promise Init()
 			CheckMicroserviceStatus();
 			ConnectToLogs();
 			LogVerbose("Completed");
+			await Migrate();
```
gabrielbeamable, line 99: I think that would be nice to have a way to tell that the migration already ran for this user, having it running only once. Maybe it could be a simple `PlayerPrefs` save variable?

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -95,6 +96,105 @@ public async Promise Init()
...
+		public async Promise Migrate()
+		{
+			var commonCsProj = await MigrateCommon();
+			foreach(var microserviceDir in Directory.EnumerateDirectories("Assets/Beamable/Microservices"))
+			{
+				var microserviceName = Path.GetFileName(microserviceDir);
+				var path = $"{StandaloneMicroservicesPath}{microserviceName}/";
+				if (!Directory.Exists(path))
+				{
+					Debug.Log(microserviceName);
```
gabrielbeamable, line 111: This log looks like could be improved to a `LogVerbose` with more information

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -95,6 +96,105 @@ public async Promise Init()
...
+					await CreateMicroService(microserviceName, $"{StandaloneMicroservicesPath}{microserviceName}/",
+					                         null,
+					                         true);
+				}
+
+				var signpost = _services.FirstOrDefault(s => s.name.Equals(microserviceName));
+				if(signpost == null) continue;
+				var command = $"add {signpost.CsprojFilePath} reference {commonCsProj}";
```
gabrielbeamable, line 119: When the line 112 above gets executed, wouldn't this create 2 references in the service csproj file?

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -115,7 +215,15 @@ public async Promise UpdateServicesVersions()
...
 			versions.OnStreamProjectVersionCommandResult(result =>
 			{
-				_projectVersion = result.data.packageVersions[0];
+				if (result.data.packageVersions.Length > 0)
+				{
+					_projectVersion = result.data.packageVersions[0];
+				}
+				else
+				{
+					Debug.Log(JsonUtility.ToJson(result));
```
gabrielbeamable, line 224: the same here for this log

client/Packages/com.beamable/Editor/BeamCli/BeamCommand.cs
```diff
@@ -422,6 +423,10 @@ public async Promise Run()
 
 					if (_exitCode != 0)
 					{
+						foreach (var err in _errors)
+						{
+							BeamEditorContext.Default.Dispatcher.Schedule(() => Debug.LogError(err.message));
```
gabrielbeamable, line 428: nice!

gabrielbeamable (changes requested):
Great! Just a few questions before approving

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
Leinnan: Good call

cdhanna (approved):
I approve this message.

gabrielbeamable (approved):

Issue #3164, opened 2024-03-13
User is always prompted with multiple warning logs when renaming content


gabrielbeamable:
Description

Renaming content causes the user to be prompted with unlimited warning logs.

Note/s:

Warning logs will be removed when launching editor but when the user decides to rename again, the bug will re-occur everytime

 

Build Details:

0.0.0-PREVIEW.NIGHTLY-202401031817

2022.3.10f1

 

Repro Steps:

Launch Editor

Install Beamable Nightly

Login/Register

Go to Content Manager

Rename a content

Result:

User is prompted with a lot of warning logs and its not stopping unless the user relaunches editor

 

Expected Result:

No Warning or Error logs when renaming content

![Image](https://github.com/beamable/BeamableProduct/assets/141041673/05aeed2b-96a5-4296-ac31-9d2083197bb9)

PR #3165, opened 2024-03-13
Unity SDK - Fix renaming content issue

gabrielbeamable:
# Ticket

This fixes issue #3164 

# Brief Description

When we were finishing writing the new name for that content, we were calling `_nameTextField.Blur();`, which was calling
`CommitName()` again, thus creating a infinite loop that sometimes was triggering a asset changing timestamp issue in Unity. I did some tests with this and wasn't able to reproduce the error, so I think this is the solution, but is hard to tell that this fixes it for sure since could have other sources as well.

# Checklist

# Notes

### DIFF
--- a/client/Packages/com.beamable/CHANGELOG.md
+++ b/client/Packages/com.beamable/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Json.Serialize` will treat non `IDictionary` values of `IEnumerable` as json arrays.
 - Disabling Content inspectors no longer causes compiler errors.
 - Leaderboard `rankgt` field is not null when specifying outlier.
+- Fix renaming content throwing infinite warnings
 
 ### Removed
 


Leinnan (approved):

client/Packages/com.beamable/Editor/UI/Content/Components/ContentVisualElement/ContentVisualElement.cs
```diff
@@ -223,12 +224,13 @@ private void NameLabel_OnKeydown(KeyDownEvent evt)
...
 		private void NameLabel_OnKeyup(KeyUpEvent evt)
 		{
+			evt.StopPropagation();
 			CheckName();
 		}
 
 		private void NameLabel_OnBlur(BlurEvent evt)
 		{
-			BeamEditorContext.Default.Dispatcher.Schedule(CommitName);
+			evt.StopPropagation();
```
cdhanna, line 233: does content renaming still work, though? If I click on a content name field, change the text, and then click on something else in Unity- how does the name get saved?

cdhanna (changes requested):

client/Packages/com.beamable/Editor/UI/Content/Components/ContentVisualElement/ContentVisualElement.cs
gabrielbeamable: Yes, the visualization in the content manager and the scriptable object of that content are both changed. Is there somewhere else that I should also check?

cdhanna: hmm, okay that seems good enough

cdhanna (approved):

PR #3166, opened 2024-03-13
fixes 3160 - validate input/output types of CLI contracts are safe to use

cdhanna:
# Ticket

Issue 3160

# Brief Description

This adds a unit test that will scan over the commands' input and output types, and make sure that, 
1. the type exists in the CLI assembly, or
2. the type is an acceptable system primitive, or
3. the type is marked with `[CliContractType]`. 

The `[CliContractType]` is the way to share C# code between the common lib and the CLI, and should only be used for things like semantic types. It shouldn't be used for network DTO objects, and so this PR also changes to some commands that were failing the test to use their own defined versions of the DTOs. 

I think I also introduced a bug with the ConstructVersionCommand, because when I generated the client code, it broke. I fixed that generation issue by adding a `ToString()` to the parameter passing.  

When the test fails, it'd look like this... Imagine the `beam me` command took a Vector2 as input for some reason...
```
Command=[AccountMeCommand] references input Type=[Vector2], which not declared in the CLI Assembly.
The CLI's output types should be defined in the CLI assembly to reduce the likelihood of accidental type changes. If the output type is changed, then
that qualifies as a potential breaking change in the CLI. 
```

# Checklist

### DIFF
--- a/cli/cli/Services/UnityCliGenerator.cs
+++ b/cli/cli/Services/UnityCliGenerator.cs
@@ -69,7 +69,7 @@ public class UnityCliGenerator : ICliGenerator
 		return files;
 	}
 
-	public static List<Type> RecurseTypes(IEnumerable<Type> inputTypes)
+	public static List<Type> RecurseTypes(IEnumerable<Type> inputTypes, bool includeTypesFromInvalidAssembly=false)
 	{
 		var output = new HashSet<Type>();
 		var toExplore = new Queue<Type>();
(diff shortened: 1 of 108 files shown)


cli/tests/BasicTests.cs
```diff
@@ -65,21 +65,22 @@ void CheckNaming(Command command, string commandName, string description, string
 			var match = Regex.Match(valueToCheck, KEBAB_CASE_PATTERN);
 			Assert.AreEqual(match.Success, true, $"{valueToCheck} does not match kebab case naming.");
 		}
-		var commandTypes = Assembly.GetAssembly(typeof(App))!.GetTypes()
-			.Where(myType => myType.IsClass && !myType.IsAbstract && myType.IsSubclassOf(typeof(Command)));
-		var commandsList = new List<Command>();
+		// var commandTypes = Assembly.GetAssembly(typeof(App))!.GetTypes()
```
gabrielbeamable, line 68: I think these comments could be removed

cli/tests/BasicTests.cs
```diff
@@ -65,21 +65,22 @@ void CheckNaming(Command command, string commandName, string description, string
...
-		{
-			var command = app.CommandProvider.GetService(type);
-			if (command != null)
-			{
-				commandsList.Add((Command)command);
-			}
-		}
+		var commandsList = app.InstantiateAllCommands();
+
+		// foreach (Type type in commandTypes)
```
gabrielbeamable, line 76: here too

cli/cli/Services/UnityCliGenerator.cs
```diff
@@ -69,7 +69,7 @@ public List<GeneratedFileDescriptor> Generate(CliGeneratorContext context)
 		return files;
 	}
 
-	public static List<Type> RecurseTypes(IEnumerable<Type> inputTypes)
+	public static List<Type> RecurseTypes(IEnumerable<Type> inputTypes, bool includeTypesFromInvalidAssembly=false)
```
gabrielbeamable, line 72: Where is this method being called with the new `includeTypesFromInvalidAssembly ` parameter? Couldn't find it, is this being called by reflection or something like that?

gabrielbeamable: I saw it in the test, but in this question I mean in the actual generation for Unity.

gabrielbeamable (changes requested):

cli/cli/Services/UnityCliGenerator.cs
cdhanna: see `UnityCliGenerator` line 44

gabrielbeamable: I see, but that is not sending the new parameter, right? So we are adding this just as an option, that we are not using right now, right?

PR #3183, opened 2024-03-15
fixes 3171 - usam project versioning

cdhanna:
a few things-

the `beam project version --requested-version` command wasn't doing what it should have been. It was loosing the `.sln` file path, and then silently failing. I believe I fixed it, and I was able to verify by creating a SAM, and then changing its nuget versions by doing `beam project version --requested-version 1.19.14`. 

Also, I removed the `_projectVersion` field and replaced it with a derived method that uses the SDK version number to come up with a nuget version number. For now, they are usually the same, except if we are developing, then we expect to use 0.0.123, which is our local nuget package, created via `set-packages.sh`. 

I made sure to have the local templates get installed through running `set-packages.sh` as well.

### DIFF
--- a/client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
+++ b/client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
@@ -36,7 +36,6 @@ namespace Beamable.Server.Editor.Usam
 		private List<BeamServiceSignpost> _services;
 		private List<BeamStorageSignpost> _storages;
 		private List<Promise> _logsCommands = new List<Promise>();
-		private string _projectVersion;
 
 		private const string BEAMABLE_PATH = "Assets/Beamable/";
 		private const string MICROSERVICE_DLL_PATH = "bin/Debug/net6.0"; // is this true for all platforms and dotnet installations?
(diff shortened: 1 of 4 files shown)


client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -99,24 +98,11 @@ public async Promise Init()
...
-				return;
-			}
-
-			var versions = _cli.ProjectVersion(new ProjectVersionArgs { requestedVersion = version?.version });
+			var nugetVersion = GetCurrentNugetVersion();
+			var versions = _cli.ProjectVersion(new ProjectVersionArgs { requestedVersion = nugetVersion });
 			versions.OnStreamProjectVersionCommandResult(result =>
 			{
-				_projectVersion = result.data.packageVersions[0];
-				LogVerbose($"Versions updated: {_projectVersion}");
```
gabrielbeamable, line 119: I believe this is going to conflict with Peters migration PR, might want to merge his PR first

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -711,7 +712,7 @@ private async Promise CreateMicroService(string serviceName, string outputPath,
 				quiet = true,
 				name = service,
 				output = outputPath,
-				version = _projectVersion
+				version = GetCurrentNugetVersion()
```
gabrielbeamable, line 715: nice, thanks for cleaning my previous mistake o/

gabrielbeamable (approved):

PR #3187, opened 2024-03-18
Issue/3172 - Add, list and remove dependencies commands in the CLI

### LINKED ISSUE #3172: CLI - project dependency commands
We need a way to list, add, and remove project dependencies via the CLI. 
If there is a Microservice, A, that depends on Microstorage, B, then

this command should list the dependencies of A. If we don't pass the service option, it should list all the dependencies grouped by service, similar to how other project commands work.
```
beam project deps list --service A
```

this command needs to add the dependency if it doesn't exist.
```
beam project deps add A B
```

and this should remove it
```
beam project deps remove A B
```

gabrielbeamable:
# Ticket

resolves #3172 

# Brief Description

Adding the following commands to the CLI:
`beam project deps list` - lists all dependencies of all services

`beam project deps list --service <name>` - lists all dependencies of the given service

`beam project deps add <microservice> <storage>` - add the <storage> as reference of the <microservice>

`beam project deps remove <microservice> <storage>` - removes the <storage> as reference of the <microservice>

# Notes

### DIFF
--- a/cli/cli/Commands/Project/Deps/DepsCommand.cs
+++ b/cli/cli/Commands/Project/Deps/DepsCommand.cs
@@ -1,0 +1,25 @@
+using Microsoft.Diagnostics.Tracing.Parsers.MicrosoftWindowsTCPIP;
+
+namespace cli.Commands.Project.Deps;
+
+public class DepsCommandArgs : CommandArgs
+{
+	
+}
+
+public class DepsCommand : AppCommand<DepsCommandArgs>, IEmptyResult
+{
+	public DepsCommand() : base("deps", "Allow access to dependencies related commands")
+	{
+	}
+
+	public override void Configure()
+	{
+	}
+
+	public override Task Handle(DepsCommandArgs args)
+	{
+		
+		return Task.CompletedTask;
+	}
+}
--- a/cli/cli/Commands/Project/Deps/DepsCommand.cs
+++ b/cli/cli/Commands/Project/Deps/DepsCommand.cs
@@ -1,4 +1,6 @@
+using Beamable.Server;
 using Microsoft.Diagnostics.Tracing.Parsers.MicrosoftWindowsTCPIP;
+using System.CommandLine.Help;
 
 namespace cli.Commands.Project.Deps;
 
@@ -19,7 +21,8 @@ public class DepsCommand : AppCommand<DepsCommandArgs>, IEmptyResult
 
 	public override Task Handle(DepsCommandArgs args)
 	{
-		
+		var helpBuilder = args.Provider.GetService<HelpBuilder>();
+		helpBuilder.Write(this, Console.Error);
 		return Task.CompletedTask;
 	}
 }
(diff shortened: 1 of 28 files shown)


cli/cli/Commands/Project/Deps/DepsCommand.cs
```diff
@@ -0,0 +1,25 @@
+using Microsoft.Diagnostics.Tracing.Parsers.MicrosoftWindowsTCPIP;
+
+namespace cli.Commands.Project.Deps;
+
+public class DepsCommandArgs : CommandArgs
+{
+	
+}
+
+public class DepsCommand : AppCommand<DepsCommandArgs>, IEmptyResult
```
cdhanna, line 10: little nitpick, we have a `CommandGroup` type that can be used, so this could be similar to
```csharp

public class ProfilingCommand : CommandGroup, IStandaloneCommand
{
	public override bool IsForInternalUse => true;

	public ProfilingCommand() : base("profile", "Commands for inspecting profiling reports")
	{
	}

}

```

cdhanna: although, we don't want the `IsForInternalUse` flag in this case.

cli/cli/Services/BeamoLocalSystem.cs
```diff
@@ -136,6 +136,52 @@ public async Task<List<string>> GetDependencies(string beamoServiceId, string pr
...
+		{
+			throw new CliException($"Failed to remove project dependency, output of \"dotnet {command}\": {result}");
+		}
+		
+		var service = BeamoManifest.HttpMicroserviceLocalProtocols[project.BeamoId];
+		var dockerfilePath = service.RelativeDockerfilePath;
+		dockerfilePath = _configService.GetFullPath(Path.Combine(service.DockerBuildContextPath, dockerfilePath));
+		var dockerfileText = await File.ReadAllTextAsync(dockerfilePath);
+		
+		string search = @$"WORKDIR /subsrc/{dependency.BeamoId}
```
cdhanna, line 169: can you explain this a bit? We are looking into the dockerfile for this line specifically, and killing it if it exists... 

I think we may want to re-think this a bit. 
I don't love the idea of having the add/remove commands just appending/removing lines arbitrarily from the midst of a dockerfile. Rather, I think we should define a section in the Dockerfile, and wholistically recompute the value of that section based on the known deps everytime. 

So instead of removing this one reference, it would be more like we are re-writing the entire deps section of the Dockerfile. I think that would represent a more stable approach.

cdhanna (changes requested):

cli/cli/Services/BeamoLocalSystem.cs
gabrielbeamable: Right now, when adding a dependency, we are looking for this tag
`# <BEAM-CLI-INSERT-FLAG:COPY> do not delete this line. It is used by the beam CLI to insert custom actions`

And adding the copy reference project part in there when adding it. 

Your suggestion is that we add a new tag just for dependencies, and every time we add/remove, we update it all using that tag to find where to write/delete, right?

cdhanna (approved):

PR #3206, opened 2024-03-21
[issue/3145] Add the assembly reference to the USAM csproj file

### LINKED ISSUE #3145: USAM - Reference Unity Assemblies
https://disruptorbeam.atlassian.net/browse/BEAM-4118

Right now you can add references to assemblies in the microservice signpost file and that will generate a csproj file for that reference which points to all source code files and Unity references. However, for the microservice to have access to those references, we need to reference that generated csproj in the microservice csprojfile.

gabrielbeamable:
# Ticket

resolves #3145 

# Brief Description

This is basically referencing in the microservice csproj file the generated project of the assemblies of that service. This happens everytime the signpost of that service is changed with new/removed assembly references.

# Notes


client/Packages/com.beamable.server/Editor/Usam/Editor/Signposts/BeamServiceAsset.cs
```diff
@@ -7,5 +10,43 @@ namespace Beamable.Server.Editor.Usam
 	public class BeamServiceAsset : ScriptableObject
 	{
 		public BeamServiceSignpost data;
+
+		private void OnValidate()
```
gabrielbeamable, line 14: This is something that I'm not sure it's what we want. The ideal world would be that we would do this after we generate the csproj files of the assembly references, however this is state dependent, due to the use of the `DotnetService`, and the `SolutionPostProcessor` thing is static. I tried calling this before this part, but it yields an error:
```
AssemblyUtil.Reload();
CsharpProjectUtil.GenerateAllReferencedAssemblies();
```

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -290,6 +290,52 @@ public async Promise CreateService(string name, ServiceType type, List<IBeamoSer
...
 
+		
+		public async Promise UpdateServiceReferences(string serviceName, List<string> assemblyReferencesNames)
+		{
+			LogVerbose($"Starting updating references");
+			//get a list of all references of that service
+			var service = _services.FirstOrDefault(s => s.name == serviceName);
+			if (service == null)
+			{
+				return;
```
cdhanna, line 301: nitpick; but this pattern rubs me the wrong way; a silent return makes the eventual bug, "I passed an invalid serviceName", really really hard to find. Can we put a warning in here, or throw an exception? If it is _never_ okay to call this method with an invalid name, then we _shouldn't_ call it with an invalid name, and if we _are_, we should KNOW about it so we can fix it.

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -290,6 +290,52 @@ public async Promise CreateService(string name, ServiceType type, List<IBeamoSer
...
+
+			
+			//remove all generated projs references
+			LogVerbose($"Removing all references from service: {serviceName}");
+			foreach (string reference in existinReferences)
+			{
+				var referenceName = Path.GetFileNameWithoutExtension(reference).Replace(CsharpProjectUtil.PROJECT_NAME_PREFIX, String.Empty);
+				var refPathToRemove = CsharpProjectUtil.GenerateCsharpProjectFilename(referenceName);
+				LogVerbose($"Removing reference: {refPathToRemove}");
+				await _dotnetService.Run($"remove {service.CsprojPath} reference {refPathToRemove}");
```
cdhanna, line 320: is it possible to add these references in parallel? 
Like, instead of `await`'ing each of them in sequence, can we start all invocations, and await the `Promise.Sequence` of them?

Same question for the later addition phase.

client/Packages/com.beamable.server/Editor/Usam/Editor/Signposts/BeamServiceAsset.cs
cdhanna: yeah the more I've worked with `OnValidate` functions, the more I hate them. They always seem to come and bite us in the butt due to AssetDatabase quirks on when they run. 

Can you remove this, and create a follow on Issue that describes how we need to be able to invoke this based on
1. the user "saving" their intention in the Inspector, 
2. when a sln file is modified.

client/Packages/com.beamable.server/Editor/Usam/Editor/Signposts/BeamServiceSignpost.cs
```diff
@@ -14,7 +15,7 @@ public class BeamServiceSignpost : ISignpostData
 		public string name;
 		public string assetProjectPath = "";
 
-		public string[] assemblyReferences;
+		public AssemblyDefinitionAsset[] assemblyReferences;
```
cdhanna, line 18: out of curiosity- what does this look like in the actual file content?

cdhanna (changes requested):
overall, looks promising! I have a few pieces of feedback

client/Packages/com.beamable.server/Editor/Usam/Editor/Signposts/BeamServiceSignpost.cs
gabrielbeamable: image
`AStar` is an assembly reference of a A* path finding asset that I downloaded from the asset store.

client/Packages/com.beamable.server/Editor/Usam/Editor/Signposts/BeamServiceAsset.cs
gabrielbeamable: Just removed the `OnValidate` method, but I left the method that I made to validate the data in the asset. I think it will be useful for the saving the intention part.

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
gabrielbeamable: agreed, just changed it for an `ArgumentException` instead

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
gabrielbeamable: It is! Thanks for pointing this out

cdhanna (approved):

PR #3208, opened 2024-03-22
run cli tests on windows and only dotnet 6

cdhanna:


cli/cli/Services/ConfigService.cs
```diff
@@ -73,20 +73,27 @@ public bool IsPathInWorkingDirectory(string path)
 	/// <returns></returns>
 	public string GetRelativePath(string relativePath)
```
gabrielbeamable, line 74: I didn't realize this in the other PR, but it's weird that the method name is `GetRelativePath` and you pass a `relativePath` to it. Maybe the name of the method could be more descriptive of what it's doing? Which is a relative specifically to an internal folder of the project

gabrielbeamable (changes requested):
Nice! Just one minor thing

gabrielbeamable (approved):

PedroRauizBeamable (approved):

PR #3209, opened 2024-03-26
[Issue/3174] Discover storages status running on docker

### LINKED ISSUE #3174: USAM - project ps command needs to list storages as well
If you have a service that depends on a storage, and you run the service rom Unity, then we need to see that the storage becomes enabled. 
Ideally, this should happen via the `beam project ps --watch` command that monitors for service run/stops.

gabrielbeamable:
# Ticket

resolves #3174 

# Brief Description

The docker service discovery was not working for storages.

# Checklist

# Notes


cli/cli/Services/DiscoveryService.cs
```diff
@@ -43,25 +43,51 @@ public async IAsyncEnumerable<ServiceDiscoveryEvent> StartDiscovery(TimeSpan tim
...
+			}
+
+			var service = new ServiceDiscoveryEntry()
 			{
 				cid = _appContext.Cid,
 				pid = _appContext.Pid,
 				prefix = MachineHelper.GetUniqueDeviceId(),
 				serviceName = serviceDefinition.BeamoId,
-				healthPort = Convert.ToInt32(protocol.HealthCheckPort),
+				healthPort = healthPort,
```
cdhanna, line 85: I don't love that the `healthPort` is being used to describe the _data_ port for a service instance. 

I think I'd prefer to use more of a union type approach, where the `ServiceDiscoveryEntry` gets two new fields, 
```
public string serviceType;
public int dataPort;
```

And the `serviceType` gets set to either `service` or `storage`, and then depending on that, either the `healthPort` OR the `dataPort` gets set.

client/Packages/com.beamable.server/Editor/Usam/Editor/CsharpProjectUtil.cs
```diff
@@ -146,7 +146,6 @@ static IEnumerable<string> GetValidDllReferences(Assembly assembly)
 
 		static IEnumerable<Assembly> GetValidAssemblyReferences(Assembly assembly)
 		{
-			var invalidPrefixes = new string[] { "Unity.", "UnityEditor.", "UnityEngine." };
```
cdhanna, line 149: did you mean to delete this? Won't this start generating invalid csproj files?

cli/cli/Services/BeamoLocalSystem_LocalRuntime.cs
```diff
@@ -145,6 +145,28 @@ public async Task StartListeningToDockerRaw(Action<string, string, Message> onSe
...
+			if (!string.IsNullOrEmpty(beamoId))
+			{
+				var beamoServiceInstance =
+					BeamoRuntime.ExistingLocalServiceInstances.FirstOrDefault(si => si.BeamoId == beamoId);
+				
+				if (beamoServiceInstance != null)
+				{
+					beamoServiceInstance.IsRunning = false;
+					var message = new Message() { ID = info.ID };
+					onServiceContainerStateChange?.Invoke(beamoId, "start", message);
```
cdhanna, line 165: is the purpose of this to simply kick-off the current state of the world?
I'm a bit worried that adding this here could have bad side-effects on other parts of the codebase that use this function (`StartListeningToDockerRaw`).

Might it make more sense to add this pre-pass step as a one-off util function in the ps, or a function you can call to emit the current running docker containers?

cdhanna (changes requested):
pretty good; a few comments

cli/cli/Services/BeamoLocalSystem_LocalRuntime.cs
gabrielbeamable: The only usage that actually passes the `onServiceContainerStateChange ` callback is the `DiscoveryService`. This is basically to inform the state, it doesn't change or add any data. Can you elaborate more on the possible side-effect of this change?

I think this problem boils down to: When I call a method to listen to a state, shouldn't it show it's current state? I went with this logic because the `beam project ps` already works like this for non-docker running services. But maybe that is something that I should keep only in the `StartDiscovery` scope of it. I saying all of this because I want to understand the reasoning behind it before just changing it 😅

cli/cli/Services/DiscoveryService.cs
gabrielbeamable: will do!

client/Packages/com.beamable.server/Editor/Usam/Editor/CsharpProjectUtil.cs
gabrielbeamable: Oh, it was an unused variable, the `IsValidReference ` already does that check ando also has these prefixes values.

cli/cli/Services/BeamoLocalSystem_LocalRuntime.cs
cdhanna: okay thats good to know that this is an additive feature. 
I don't like the idea of using a callback to initiate start-state though. To me, that feels like the caller should WAIT for the start state instead of using callbacks to get the data. 

I guess mostly I'm advocating for refactoring this into two methods,
1. get the start state without a callback, (using await)
2. set up the listener for actual asyncrounous state changes

gabrielbeamable:
> pretty good; a few comments

comments addressed!

cdhanna (approved):

PR #3224, opened 2024-03-28
fixes 3215 - cli installs projects for current service version

cdhanna:
Now, when you do
```sh
beam project new service TunaEye
```

you'll...
1. be creating a service with the version for the _current_ CLI install
2. not be prompted to install templates (they'll install automatically)

If you're running a local dev version, you'll get 0.0.123 as the version, which is the nuget version from `./set-packages.sh`

If you pass in an explicit `--version`, we'll use that, but it must be a valid package.

TO update the CLI itself, use `beam version install` or `beam version update` (i added an alias)


cdhanna:
backporting this to 1.19.x is going to be a chore.

gabrielbeamable (approved):

Issue #3190, opened 2024-03-18
beam project generate-env should not write a file anymore


cdhanna:
It is a security risk to be writing the .env file into the bin folder for execution.
The env file is getting generated by the service itself, so we could just pass the information back on STD OUT instead. 

In the `Prepare<TMicroservice>` method of the `MicroserviceBootstrapper`, we call `beam generate-env`
That generates a `.env` file that contains the realm secret, which can get left on disk and pose a security-at-rest concern.

```sh
chrishanna@Chriss-MacBook-Pro-2 net6.0 % cat .env 
SECRET=<oops!>
HOST=wss://api.beamable.com/socket
CID=1338004997867618
PID=DE_1650190063990785
NAME_PREFIX=Chriss-MacBook-Pro-2
BEAM_INSTANCE_COUNT=1
````

Instead, we should change the `generate-env` command to output the env data over the json piping protocol, and the`Prepare` method to read that and set the local environment instead of reading the file off disk.

We should remove the function that reads the file, 
```csharp
		static void LoadEnvironmentVariables(string filePath=".env")
		{
			if (!File.Exists(filePath))
				throw new Exception($"No environment file found at path=[{filePath}]");

			foreach (var line in File.ReadAllLines(filePath))
			{
				var parts = line.Split(
					'=',
					StringSplitOptions.RemoveEmptyEntries);

				if (parts.Length != 2)
					continue;

				Environment.SetEnvironmentVariable(parts[0], parts[1]);
			}
		}
```

and put that responsibility into the `Prepare` method.

**We should write a blank .env file**
this way, old .env files will get replaced with a blank file, and also it makes sure that upgraders don't need to manually edit their `Program.cs`
When we write the blank .env file, we need a comment that explains WHY we are doing that, for legacy purposes.

We can also mark this command as internal.

PR #3228, opened 2024-04-01
fixes 3190 - generate env uses stdout instead of file

cdhanna:
#3190


cli/cli/Commands/Project/GenerateEnvFileCommand.cs
```diff
@@ -36,30 +41,26 @@ public override void Configure()
...
-BEAM_INSTANCE_COUNT={args.instanceCount}
-REFRESH_TOKEN={refreshToken}
-";
-
-
+		var output = new GenerateEnvFileOutput
+		{
+			envVars = new List<EnvVarOutput>
+			{
+				EnvVarOutput.Create("SECRET", proj.secret),
```
gabrielbeamable, line 54: Is there any possibility for these environment variables conflict with different ones in that process we are adding them? If so, would be nice to add a prefix or something to these.

gabrielbeamable (changes requested):
Just a small concern that can be nothing

cli/cli/Commands/Project/GenerateEnvFileCommand.cs
cdhanna: fair enough feedback, but I think that would amount to a larger breaking change than we should do in a patch release. Do you mind if I make a ticket that gets on a 2.0 milestone?

gabrielbeamable: seems good enough

gabrielbeamable (approved):

cli/cli/Commands/Project/GenerateEnvFileCommand.cs
cdhanna: https://github.com/beamable/BeamableProduct/issues/3231

PR #3264, opened 2024-04-04
adds spew logging for cli

cdhanna:
Now, if you enable SPEW_ALL or SPEW_CLI, you'll get very verbose CLI logging for every command invoked, and all stdout and stderr from each command.


gabrielbeamable (approved):
Nice!

PR #3282, opened 2024-04-09
[issue/3268] Commit and reference microservice assemblies

### LINKED ISSUE #3268: USAM - Call for services assembly reference update
Right now we have a method in the `CodeService` that goes through the list of assembly references names of a service and update those in that service csproj file. The method name is `UpdateServiceReferences`. What we need to do is to call this method in the right moment, which would be right after the `SolutionPostProcessor.OnPreGeneratingCSProjectFiles` finished. For this to work, we can:

1. Save the desired list of references that are in the signpost when there is a modification in it. In this step is important to validate that those references can be set. There is a `CheckAllValidAssemblies()` method in the `BeamServiceAsset` that can be used for this purpose. The `CodeService` itself has access to the signpost files, so this would be more about not letting wrong information get written in the asset when a change is committed.
2. Call the `UpdateServiceReferences` after the `SolutionPostProcessor.OnPreGeneratingCSProjectFiles`. This is the biggest challenge here because the `OnPreGeneratingCSProjectFiles` method is static and right now, in order to run dotnet commands, we use a service instance named `DotnetService`, which is not stateless.

gabrielbeamable:
# Ticket

resolves #3268 

# Brief Description

This creates a button that checks if the assembly references of a service are valid, and if so it references it in that service
csproj file.

# Notes

### DIFF
--- a/client/Packages/com.beamable.server/Editor/Usam/Editor/Signposts/BeamServiceImporter.cs
+++ b/client/Packages/com.beamable.server/Editor/Usam/Editor/Signposts/BeamServiceImporter.cs
@@ -84,9 +84,9 @@ namespace Beamable.Server.Editor.Usam
 				{
 					if (_serviceAsset != null)
 					{
-						if (!_serviceAsset.data.CheckAllValidAssemblies())
+						if (!_serviceAsset.data.CheckAllValidAssemblies(out string message))
 						{
-							Debug.LogError("Assembly references in file are not valid");
+							Debug.LogError($"Error: {message}");
 						}
 						else
 						{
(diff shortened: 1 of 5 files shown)


client/Packages/com.beamable.server/Editor/Usam/Editor/Signposts/BeamServiceImporter.cs
```diff
@@ -79,6 +79,23 @@ public override void OnInspectorGUI()
...
 
 				}
+				
+				if (GUILayout.Button("Apply Changes"))
+				{
+					if (_serviceAsset != null)
+					{
+						if (!_serviceAsset.data.CheckAllValidAssemblies())
+						{
+							Debug.LogError("Assembly references in file are not valid");
```
cdhanna, line 89: it would be good to say WHY the file isn't valid. In the `CheckAllValidAssemblies` function, I see you're returning false in a few places... what about changing the sig to something like, 

```csharp
public static bool CheckAllValidAssemblies(out string validationMessage) {
  validationMessage = null;
  if (notValidSomeHow) { 
    validationMessage = "the eclipse flipped all the bits";
    return false;
  }
  return true;
}

cdhanna (changes requested):
one q

cdhanna (approved):

PR #3290, opened 2024-04-11
runtime ssl fix

cdhanna:


cdhanna:
image

allister-beamable (approved):
The retry logic looks sound.

Is there any other possible place that this SSL-oriented retry logic would need to be implemented, or are we sure that this and the previous (editor) one are the only two places?

PR #3324, opened 2024-04-22
[Issue/3269] USAM migration progress and other fixes

### LINKED ISSUE #3269: USAM - Migration Confirmation Popup with progress info
Something similar to what the Download Content popup does, in which updates the state of download of each content it has. We could do the same for each service/storage that is going to be migrated, checking those completed already and showing a success confirmation at the end of the process.

Each microservice has granular progress, so the ENTIRE loading bar should be the combination of every microservice's granular progress.

We need to find a way to delay the domain reload until after all microesrvices have been migrated. The domain reload happens because we re-init the `CodeService` at the end of the Microservice. So we need a way to delay that. Could be as simple as a new method arg.

There are unknowns when it comes to UI and how to handle loading bars and such.

gabrielbeamable:
# Ticket

resolves #3269 

# Brief Description

The migration popup now has a progress info of what is being done. Also, the migration happens asynchronously and dependencies between services and storages are being migrated as well. 

# Notes


cdhanna (approved):

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -91,10 +91,6 @@ public async Promise Init()
 			LogVerbose("Saving all libraries referenced by services");
 			await SaveReferencedLibraries();
 
-			LogVerbose("Set manifest start");
-			await SetManifest(_cli, _services, _storages);
```
gabrielbeamable, line 95: This was duplicated for some reason, probably from some conflict resolution

client/Packages/com.beamable.server/Editor/Usam/Editor/MigrationHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using UnityEditorInternal;
+
+namespace Beamable.Server.Editor.Usam
+{
+	public static class MigrationHelper
+	{
+
+		public static List<string> GetDependentServices(string storageName)
```
gabrielbeamable, line 11: All methods from this file except from this one were copied from the editor popup that is used in the old microservice manager to show dependencies

PR #3338, opened 2024-04-23
fixes 3326

cdhanna:
fixes 3326 by not throwing exception if source doesn't exist.


gabrielbeamable (approved):

microservice/microserviceTests/CronBuilderTests/JobConversionTests.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+	[TestCase]
+	public void CanDeserializeJobWithoutSource()
+	{
+		var job = new JobDefinition
+		{
+			id = "test",
+			name = "test",
+			owner = "test",
+			source = new OptionalString(),
```
allister-beamable, line 18: The semantics here are that `new OptionalString()` with no constructor argument is an empty optional string? That is, `new OptionalString()` is the equivalent of Scala`s `None`?

client/Packages/com.beamable.server/CHANGELOG.md
```diff
@@ -6,12 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [1.19.18]
+### Fixed
+- `BeamScheduler` can deserialize jobs without a `source` field.
+- `AssumeNewUser` does not allow `userId` that is not a positive value
```
allister-beamable, line 11: This seems not-relevant to the current changes.

allister-beamable (changes requested):
Code looks good, but the changelog appears to have a rogue entry about AssumeNewUser.

allister-beamable (approved):
I was wrong about the rogue changelog entry. The AssumeNewUser line was already there, it just moved along with the `### Fixed` header.

microservice/microserviceTests/CronBuilderTests/JobConversionTests.cs
cdhanna: correct- that is a `None`

PR #3342, opened 2024-04-23
[issue/3249] AssumeNewUser should not allow negative userId

### LINKED ISSUE #3249: AssumeUser should forbid invalid player IDs, especially forbidding 0 as the playerId
**Describe the bug**
When using `AssumeUser` in C# microservice code, if the `playerId` argument is 0, this can lead to problems when using APIs that involve in-flight messages, such as for example Groups service join/leave requests, the backend encounters errors trying to look up the user whose player ID is 0. As such, `playerId`=0 should be avoided categorically.

**To Reproduce**
```csharp
[ClientCallable, SwaggerCategory("GroupsTests")]
public async Promise<string> TestAssumeUserLeaveGroup(long playerId, long groupId)
{
	var assumed = AssumeUser(playerId, false);
	try
	{
		await assumed.Services.Groups.LeaveGroup(groupId);
		return $"Succeeded with assumed user: {assumed.Context.UserId}";
	}
	catch (Exception e)
	{
		return $"Failed with assumed user: {assumed.Context.UserId}; exception: {e.Message}";
	}
}
```
Invoke this callable from the Swagger docs page in the portal and supply `"playerId": 0`

**Expected behavior**
Call fails with an exception (outside the try/catch) saying that playerId 0 is not valid.

**Actual behavior**
Call succeeds (at least by virtue of the try/catch) and an in-flight message gets created for playerId 0 (only verifiable by backend database access).

**Additional context**
My test that I showed in the repro steps is purposely using the deprecated `AssumeUser` as opposed to `AssumeNewUser`, but both of them should forbid invalid player IDs.

gabrielbeamable:
# Ticket

resolves #3249 

# Brief Description

We should forbid invalid userIds in the call of `AssumeNewUser`. However that is more complicated since
it would require an API call to do so. In that case we are only checking if the `userId` is a positive value.

# Checklist

# Notes

### DIFF
--- a/client/Packages/com.beamable.server/CHANGELOG.md
+++ b/client/Packages/com.beamable.server/CHANGELOG.md
@@ -5,7 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-=======
 ## [1.19.18]
 ### Changed
 - `async void` methods are not allowed in Microservices, and will cause the Microservice to fail. Instead, consider using methods with `async Task`, or `async Promise`. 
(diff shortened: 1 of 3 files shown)


cdhanna (approved):
nice

PR #3354, opened 2024-04-26
fixes 3335 - usam refresh button broken, and some speed improvements

cdhanna:
When you click refresh, we need to re-init, not just refresh the services; this is because the refresh function relies on variables destroyed by the domain reload. 

Also, I added in the log EOF thing for CLI invocations, to allow for the SDK client to stop caring about process handles after they receive the EOF signal. 

Also also, We were doing to recursive file scans, and I combined them into one, which saved about 250ms on my machine (each recursive scan took 250ms, so cutting one was good).

### DIFF
--- a/client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
+++ b/client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
@@ -69,6 +69,9 @@ namespace Beamable.Server.Editor.Usam
 		{
 			if (EditorApplication.isPlayingOrWillChangePlaymode)
 				return;
+			
+			UsamLogger.ResetLogTimer();
+			
 			UsamLogger.Log("Running init");
 			_services = GetBeamServices();
 			UsamLogger.Log("Have services");
@@ -119,6 +122,7 @@ namespace Beamable.Server.Editor.Usam
 			
 			
 			UsamLogger.Log("Completed");
+			UsamLogger.StopLogTimer();
 		}
 
 		public async Promise Migrate(List<IDescriptor> allDescriptors, Action<float, string> updateCallback, CancellationToken token)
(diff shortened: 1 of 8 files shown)


client/Packages/com.beamable/Editor/BeamCli/BeamCommand.cs
```diff
@@ -352,22 +359,34 @@ public async Promise Run()
 							return;
 						}
 
-						_dispatcher.Schedule(() =>
+						// if (!_earlyExit)
```
gabrielbeamable, line 362: Just a commented code left in here

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -69,11 +69,12 @@ public async Promise Init()
...
 				return;
+			
+			UsamLogger.ResetLogTimer();
+			
 			UsamLogger.Log("Running init");
-			_services = GetBeamServices();
-			UsamLogger.Log("Have services");
-			_storages = GetBeamStorages();
-			UsamLogger.Log("Have storages");
+			GetBeamServicePosts(out _services, out _storages);
```
gabrielbeamable, line 76: There are more places in our Unity SDK that could use this improvement. One example is in the `OnGeneratedSlnSolution` method we call both `GetBeamServices` and `GetBeamStorage` to go through all csproj files and reference them in Unity sln. We do use these methods in other places too, so it would be nice to check if they can use this

gabrielbeamable (changes requested):
Just a couple of things

client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
cdhanna: Made a few more use cases of this method.

gabrielbeamable (approved):

PR #3362, opened 2024-05-02
[3360] Restructure the way that microservice clients are generated to make it easier for customer to choose in which module they want to use it

PedroRauizBeamable:
# Ticket

https://github.com/beamable/BeamableProduct/issues/3360

# Brief Description

Unreal Microservice Client CodeGen no longer injects the clients into an existing module; instead it creates a new plugin and module "ProjectName"MicroserviceClients and puts the clients there.

This solves 2 issues:
- Customers can reference this plugin inside other plugins from which they might want to call microservices.
  - For customers, this also enables the ability to integrate microservice calls into the OnlineSubsystemBeamable package's "Customer" sub-folders.
- We can also reference this plugin from any BEAMPROJ_ samples we have and want to access a specific microservice we have declared in our main Unreal repo.

# Checklist

# Notes


cdhanna (approved):
I'll admit this is a bit over my head, but I don't see anything bad in here.

PR #3374, opened 2024-05-09
[Issue/3270] Make docker context path be the parent of the beamable config folder

### LINKED ISSUE #3270: CLI - services run not aways working
The command `beam services run` is not working for standalone microservices that doesn't have the following folder structure:

TestService/
----services/
       ----TestService/
       ----TestServiceCommon/
----TestService.sln

This seems to be because the DockerFile is trying to copy the Common project assuming it's a child of the `services/` folder.

-------

When running from Unity, we need to set the DockerBuildContextPath to the unity project root, and set the relative Dockerfile path be relative to the unity root. 
```json
"HttpMicroserviceLocalProtocols": {
    "EggSalad": {
      "DockerBuildContextPath": ".",
      "RelativeDockerfilePath": "Assets/Beamable/StandaloneMicroservices~/EggSalad/Dockerfile",
      "BindSrcForHotReloading": {
        "IsReadOnly": false,
        "LocalPath": null,
        "InContainerPath": null
      },
      "HotReloadEnabledEndpoint": null,
      "HotReloadEnabledPort": null,
      "CustomPortBindings": [],
      "CustomBindMounts": [],
      "CustomVolumes": [],
      "CustomEnvironmentVariables": [],
      "InstanceCount": 1
    },
```

The beamservice signpost file has the `assetProjectPath` field, which needs to map to the Unity project relative path of the service csproj file. Then, that path becomes the main source for the relative dockerfile path, and the build context path is always `.`, because Unity CLI .beamable contexts are always at the unity root.

gabrielbeamable:
# Ticket

resolves #3270 

# Brief Description

While working on fixing the `beam services run` command, I found that because we use standalone microservices in different ways, for example in the Unity SDK, then we were previously, it means that the way we were handling the docker context path was wrong and would only work for what is inside the microservice folder, and not work with dependencies for example. This basically fixes that, making the docker context path be the parent folder of the beamable config folder in the project.

# Notes


cli/cli/Services/BeamoLocalSystem.cs
```diff
@@ -353,9 +359,14 @@ private async Promise UpdateDockerFile(string serviceName)
 
 		foreach (var dependency in dependencies)
 		{
-			string depName = dependency.name;
-			newText = newText.Replace(endTag, replacement.Replace(serviceNameTag, depName));
+			string path = _configService.GetRelativeToBeamableFolderPath(dependency.projPath);
```
gabrielbeamable, line 362: this should be relative to the docker context path.

cli/cli/Services/BeamoLocalSystem.cs
```diff
@@ -309,11 +309,12 @@ public async Task AddProjectDependency(BeamoServiceDefinition project, BeamoServ
 	public static bool ValidateBeamoServiceId_DoesntExists(string beamoServiceId, List<BeamoServiceDefinition> serviceDefinitions) =>
 		!serviceDefinitions.Contains(new BeamoServiceDefinition() { BeamoId = beamoServiceId }, new BeamoServiceDefinition.IdEquality());
 
-	private async Promise UpdateDockerFile(string serviceName)
+	public async Promise UpdateDockerFile(BeamoServiceDefinition serviceDefinition)
```
gabrielbeamable, line 312: call this from services run and services deploy

cli/cli/Services/BeamoLocalSystem_Docker.cs
```diff
@@ -320,39 +320,79 @@ public Task<string> PullAndCreateImage(string publicImageName, Action<float> pro
...
+	{
+		var paths = new List<string>();
+
+		var fileContents = File.ReadAllLines(dockerFilePath);
+
+		foreach (string line in fileContents)
+		{
+			if (line.StartsWith("COPY"))
+			{
+				var parts = line.Split(" ");
```
gabrielbeamable, line 333: use remove empty entries here, and check if this array has COPY SOURCE DESTINATION, 3 elements, then it's correct. And throw specific exceptions that we care about in the GetAttributes call

cli/cli/Services/BeamoLocalSystem_HttpMicroservice.cs
```diff
@@ -34,6 +34,11 @@ public partial class BeamoLocalSystem
 			BeamoProtocolType.HttpMicroservice,
 			async (definition, protocol) =>
 			{
+				if (string.IsNullOrEmpty(buildContexPath))
+				{
+					buildContexPath = _configService.GetDockerBuildContextPath();
```
gabrielbeamable, line 39: remove this parameter and just get the value from this method from ConfigService

cdhanna (approved):
💣

PR #3391, opened 2024-05-16
[3390] Fixes to Deploy/Run CLI commands (adjusted templates in preparation for 2.0)

PedroRauizBeamable:
# Ticket

https://github.com/beamable/BeamableProduct/issues/3390

# Brief Description

- Fixed issue with deploy/run commands that caused the build inside the container image to fail (wasn't adding common library)
- Re-worked dockerfile to support internal Sample development and cleanup old unsupported things
- Re-worked set-packages.sh to work correctly with Windows (was already working on Mac and still is)
- Added DockerBuildArgs struct to Deploy/Run functions of BeamoLocalSystem in order to support the BUILD_MODE ARG (sample development workflow)

# Notes


cli/cli/Services/BeamoLocalSystem.cs
```diff
@@ -330,13 +340,10 @@ public async Promise UpdateDockerFile(BeamoServiceDefinition serviceDefinition)
 			"# <BEAM-CLI-INSERT-FLAG:COPY> do not delete this line. It is used by the beam CLI to insert custom actions";
 		const string serviceNameTag = "<SERVICE_NAME>";
 		const string servicePathTag = "<SERVICE_PATH>";
-		string toAddCsproj = $"COPY {servicePathTag}/{serviceNameTag}.csproj .";
```
cdhanna, line 333: 💣

client/set-packages.sh
```diff
@@ -1,10 +1,11 @@
 PROJECTS_DIR=${PROJECT_DIR_OVERRIDE:-"../ProjectsSource/"}
-HOME_FOR_BUILD=${HOME_OVERRIDE:-$HOME}
+HOME_FOR_BUILD=${1:-$HOME}
```
cdhanna, line 2: I think this may break some build scripts on GHA. The buildPr workflow needs to override these, but I guess it could be modified to pass ${{github.workspace}} in as a parameter? 
```yaml
      - name: Install local packages
        working-directory: ./client
        shell: bash
        run: |
          ./set-packages.sh
        env:
          HOME_OVERRIDE: ${{ github.workspace }}
          PROJECT_DIR_OVERRIDE: ${{ github.workspace }}/NugetBuildDir/
```

templates/BeamService/Dockerfile
```diff
@@ -1,5 +1,15 @@
-﻿# use the dotnet SDK as a build stage
-FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine AS build-env
+﻿ARG BUILD_MODE=beamable_dev
+
+# Processes specific 
+FROM mcr.microsoft.com/dotnet/sdk:6.0-alpine as build_game_maker
```
cdhanna, line 4: I am anxious about putting this in the default template for all customer microservices. I understand the value to _us_, but the value to our customers feels very low, and the clutter it costs in the dockerfile feels bad. 

I'm fine with this for now, but I wonder if there is a world where we replace this with more generated dockerfile nonsense. Like, 

```Dockerfile
# <BEAM-CLI-PROCESS> this line signals the start of custom logic injected by the CLI based upon the --build-mode property.

# </BEAM-CLI-PROCESS>
```

cdhanna (changes requested):
looks good. 
One probable issue about the buildPR workflow, and logging a sadness for the books. But otherwise, :+1:

client/set-packages.sh
cdhanna: I've attempted this fix; we'll see. If it works I'll change my review to approve.

cdhanna (approved):

PR #3394, opened 2024-05-20
fixes 3328 - make save-to-file opt out

cdhanna:
Now when you do, 
```
beam init
```

the `-save-to-file` stuff happens automatically. Actually, there is a new property, `--no-token-save` that allows you to opt-OUT of token saving. I left the ability to specify `--save-to-file=false` as a backwards compat thing. 

I also took this opportunity to change "username" to "email", because I think its a source of confusion. 

Also, the build release stuff is currently broken, so I masked it with a build-prop.

### DIFF
--- a/cli/cli/Options/SaveToFileOption.cs
+++ b/cli/cli/Options/SaveToFileOption.cs
@@ -1,6 +1,73 @@
-﻿namespace cli;
+﻿using System.CommandLine;
 
-public class SaveToFileOption : ConfigurableOptionFlag
+namespace cli;
+
+public class SaveToFileOption
 {
-	public SaveToFileOption() : base("save-to-file", "Save login refresh token to file") { }
+	/// <summary>
+	/// Add the --save-to-file and --no-token-save options to a command.
+	/// The --save-to-file existed in 1.x, and was defaulted to FALSE.
+	///
+	/// In 2.0, --no-token-save was introduced, and the intention is that tokens
+	/// DO save by default. The new option exists as a way to OPT OUT. 
+	/// </summary>
+	/// <param name="command"></param>
+	/// <typeparam name="T"></typeparam>
+	public static void Bind<T>(AppCommand<T> command)
+		where T : CommandArgs, IArgsWithSaveToFile
+	{
+		const string saveToFileStr = "--save-to-file";
+		const string noTokenSaveStr = "--no-token-save";
+
+		const bool saveToFileDefault = true;
+		const bool noTokenSaveDefault = false;
+		
+		// included for legacy reasons. 
+		var saveToFileOption = new Option<bool>(saveToFileStr, 
+			
+			// Hey You, Watch out! This default value affects the algorithm in the parse handling! 
+			getDefaultValue: () => saveToFileDefault, 
+			description: "Save login refresh token to file");
+		
+		// this option is legacy, and we don't really want people using it anymore. 
+		saveToFileOption.IsHidden = true;
+		
+		var noTokenOption = new Option<bool>(noTokenSaveStr, 
+			
+			// Hey You, Watch out! This default value affects the algorithm in the parse handling! 
+			getDefaultValue: () => noTokenSaveDefault, 
+			description: $"Prevent auth tokens from being saved to disk. This replaces the legacy {saveToFileStr} option.");
+		
+		command.AddOption<bool>(saveToFileOption, (args, value) => args.SaveToFile = value);
+		command.AddOption<bool>(noTokenOption, (args, context, noTokenSave) =>
+		{
+			var saveToFile = context.ParseResult.GetValueForOption(saveToFileOption);
+
+			switch (saveToFile, noTokenSave)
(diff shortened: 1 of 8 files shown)


gabrielbeamable (approved):

cli/cli/Options/SaveToFileOption.cs
```diff
@@ -1,6 +1,73 @@
...
+		const string saveToFileStr = "--save-to-file";
+		const string noTokenSaveStr = "--no-token-save";
+
+		const bool saveToFileDefault = true;
+		const bool noTokenSaveDefault = false;
+		
+		// included for legacy reasons. 
+		var saveToFileOption = new Option<bool>(saveToFileStr, 
+			
+			// Hey You, Watch out! This default value affects the algorithm in the parse handling!
```
allister-beamable, line 28: What algorithm? Where? Filename and approximate placement (at least which class/method it is inside) would be good.

cli/cli/Options/SaveToFileOption.cs
```diff
@@ -1,6 +1,73 @@
...
+			// Hey You, Watch out! This default value affects the algorithm in the parse handling! 
+			getDefaultValue: () => saveToFileDefault, 
+			description: "Save login refresh token to file");
+		
+		// this option is legacy, and we don't really want people using it anymore. 
+		saveToFileOption.IsHidden = true;
+		
+		var noTokenOption = new Option<bool>(noTokenSaveStr, 
+			
+			// Hey You, Watch out! This default value affects the algorithm in the parse handling!
```
allister-beamable, line 37: Same as above. Please reference the algorithm explicitly so that readers can understand it without too much hunting.

cli/cli/Options/SaveToFileOption.cs
```diff
@@ -1,6 +1,73 @@
...
+			// Hey You, Watch out! This default value affects the algorithm in the parse handling! 
+			getDefaultValue: () => noTokenSaveDefault, 
+			description: $"Prevent auth tokens from being saved to disk. This replaces the legacy {saveToFileStr} option.");
+		
+		command.AddOption<bool>(saveToFileOption, (args, value) => args.SaveToFile = value);
+		command.AddOption<bool>(noTokenOption, (args, context, noTokenSave) =>
+		{
+			var saveToFile = context.ParseResult.GetValueForOption(saveToFileOption);
+
+			switch (saveToFile, noTokenSave)
```
allister-beamable, line 46: Is this the algorithm referenced by the comments on lines 28 and 37? Why is part of it expressed as concrete `true` and `false` values but other cases are expressed in terms of the `...Default` variables? Is there any way that it could be expressed wholly in concrete boolean values?

allister-beamable (changes requested):
I am left uneasy by the "Hey you!" comments; they smell almost as bad as a TODO. If they are truly inevitable, then proceed, but their very presence worries me.

cli/cli/Options/SaveToFileOption.cs
cdhanna: Yeah, I see what you mean. 

I pulled the "algorithm" into a separate static function at least, and I removed the mix of default refs and literals, so now it is all just boolean literals. 

I'm not sure what to do about the comments, _really_. I reworded them slightly. I tried writing a version of the algorithm that would handle arbitrary defaults, but my :brain: couldn't handle thinking about all the various cases. 
The point I'm trying to make is that these default values are _important_, more so than the code-reader may originally think. The implementation of what value gets used for the question, "should I save the token?" is based on if the given config flags are set AWAY from their default.

allister-beamable: Okay. I still do not exactly _like_ it, but I am willing to accept it if there is no simpler way to achieve the same effect.

My big concern is that if it is hard to explain in code, it will be hard to explain to the humans trying to use it.

That said, it may not be worth a whole lot of philosophical thrashing over something that we expect people to rarely touch except when doing highly machine-oriented scripting. (in which case they'll be very explicit with the flags and do testing--as long as we don't suddenly change the behavior out from under those power users, all will be well)

cli/cli/Options/SaveToFileOption.cs
```diff
@@ -18,56 +18,53 @@ public static void Bind<T>(AppCommand<T> command)
...
-
-		const bool saveToFileDefault = true;
-		const bool noTokenSaveDefault = false;
 		
 		// included for legacy reasons. 
 		var saveToFileOption = new Option<bool>(saveToFileStr, 
 			
-			// Hey You, Watch out! This default value affects the algorithm in the parse handling! 
-			getDefaultValue: () => saveToFileDefault, 
+			// This default value is assumed "true" in the SaveToFileOption.ShouldSaveToFile function
```
allister-beamable, line 25: For what it is worth, this wording is _much_ nicer 😄; it is very explicit about what the assumption is and where the assumption matters. 👍

allister-beamable (approved):

PR #3444, opened 2024-05-29
fixes 3359 - adds optional serialization for C#MS

cdhanna:
So the gist is that we never had support for correctly deserializing `Optional` types in the C#MS. 

This adds a converter to properly handle the reading of those values.


microservice/microserviceTests/microserviceTests.csproj
```diff
@@ -13,6 +13,7 @@
         <OutputType>Library</OutputType>
 
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
+        <RollForward>LatestMajor</RollForward>
```
allister-beamable, line 16: What is RollForward? Is it pertinent to Optional types in JSON serialization?

allister-beamable (approved):
Looks reasonable; just one question about something that looks like a piggy-back change.

microservice/microserviceTests/microserviceTests.csproj
cdhanna: oh yeah, this is a piggy back change :+1: 

It means that if you only have net8 installed (like I do, now), the project _allows_ itself to be run even if it is targeting net6. 

In this case, the unit test project is saying, 
"hey, I am at minimum a net 6 project!"
"but also, its fine to use net7 or 8".

PR #3446, opened 2024-05-30
fixes 3445 - BeamablePackages delayCall

cdhanna:
I swear at one point, there were Unity docs advising us to ping against their Package API using `delayCall`. The docs currently advise to use `EditorApplication.update` instead, which I've done in this PR. 

Looking at profiler reports, it becomes clear that the `delayCall` approach is especially nasty, because part of the execution logic is to append a delegate to the `delayCall` action. And apparently, that gets more and more expensive the more often you do it. Perhaps because in order to combine the delegates, it needs to re-capture variables or something. 

But this approach in the PR attaches the delegate _once_, and then is responsible for clearing the delegate when the operation completes.


cdhanna:
I did run a profiler, and saw _less_ GC.

allister-beamable (commented):
As long as we are satisfied that the callback on update is getting properly removed in all circumstances that it should, this looks right and proper.

PR #3451, opened 2024-05-30
developer docs

cdhanna:


Leinnan:
I love the idea! <3 
Sidenote - would you consider putting that into some linter to make more in line with Markdown recommendations? Something like https://dlaa.me/markdownlint/

Leinnan:
Also we can consider adding `docs` folder to formatCode job: https://github.com/beamable/BeamableProduct/blob/main/.github/workflows/formatCode.yml (look for `Lint changelogs`)

PR #3485, opened 2024-06-13
fixes #3313 - global log options

### LINKED ISSUE #3313: Add global logging options to CLI
https://disruptorbeam.atlassian.net/browse/BEAM-4100

The CLI will currently mask tokens by default, and log to a file for diagnostic purposes. 

However, both of these features should be configurable by global feature flags.

 

--unmask-logs should prevent log masking, and --no-log-file should prevent using a log file sink.

cdhanna:
adds the `--unmask-logs` and `--no-log-file` option.

Some trickery is done to get access to the parse options, and essentially the log config needs to be delayed until we have the parse result from the CLI framework. 

so now if you pass `--unmask-logs`, the logs won't have the tokens set to last-4.
And if you pass `--no-log-file`, then no log file is written :shrug: 

---

Oh, also I fixed a bug with the `beam server req` command, it was using the wrong port by default :cry:


cli/cli/App.cs
```diff
@@ -607,12 +611,12 @@ protected virtual Parser GetProgram()
 				reporter.Exception(ex, context.ExitCode, context.BindingContext.ParseResult.Diagram());
 			}
 
-			var logConfigData = CommandProvider.GetService<LogConfigData>();
-			if (logConfigData.shouldUseTempFile)
+			// var appContext = CommandProvider.GetService<IAppContext>();
```
gabrielbeamable, line 614: comment leftover here

cli/cli/Options/GlobalOptions.cs
```diff
@@ -0,0 +1,29 @@
+using System.CommandLine;
+
+namespace cli.Options;
+
+
+public class UnmaskLogsOption : Option<bool>
```
gabrielbeamable, line 6: Pretty good option to have! Nice!

gabrielbeamable (approved):

Issue #3473, opened 2024-06-07
CLI - SAMs project should use local CLI install


cdhanna:
Today, a `.beamable` workspace is a single folder... But we should include a sibling folder, `.config` that includes the idiomatic dotnet local tool config, 
```json
{
  "version": 1,
  "isRoot": true,
  "tools": {
    "beamable.tools": {
      "version": "2.0.0",
      "commands": [
        "beam"
      ]
    }
  }
}
```

The global version becomes a setup only version, to install SAMs. If you aren't using SAM, this whole convo is almost meaningless. 
The local version is the source-of-truth, and is defined by `.config/dotnet-tools.json` as a sibling to the `.beamable` folder.

`dotnet-tools.json` is the source of truth for the entire project for 
1. nuget packages, 
2. cli usages

SDK integrations are responsible for setting `dotnet-tools.json`. (From Unity, it'll come from `versions-default.json`, in UE it comes from ???)

Then, we need
1. some sort of msbuild wizardry to keep the nuget package in-sync with the local tools version. Perhaps we use an msbuild property in the csproj that is auto-infered from the tool path.

If the user invokes the global CLI, 
1. at min, identify the error, and log an error.
2. better yet, auto-correct the issue and proxy to the local CLI, but log a loud warning and don't do this step on `--no-local-proxy` or something. 

-----

all beamable versions are derived from the `dotnet-tools.json` file. Pedro has added msbuild wizardry to extract an msbuild property that _is_ the version number in the json. 

At engine integration, we need to have a `.config` folder that defines the CLI

cdhanna:
the proxy of the global to local needs to be
1. loud, and announce that it is happening, and mention
2. there is a flag to disable it

PedroRauizBeamable:
PR for it is here: https://github.com/beamable/BeamableProduct/pull/3490

All tests are passing locally, so...  there's still needs some things to be done. Namely, getting GHA to play nicely with the new stuff as well as fixing any MacOS/Linux issues.

PR #3490, opened 2024-06-16
Issue/3473 - Project-Scoped CLI for Engine Integration

PedroRauizBeamable:
# Ticket
https://github.com/beamable/BeamableProduct/issues/3473

# Brief Description

## Various Features & Fixes
- [New Feature] Microservice Client Code generation through OAPI now support adding arbitrary data-types to the list of data types that are generated for the target engine.
- [Change] Modified Unreal CLI code-gen to better support the CLI running in server mode;
- [Fix] Content manifest performance improvements (we no longer download content serially, instead we do it concurrently)

## CLI Local vs Global
- Modified microservice `csproj` templates and injected props so that it keeps the Beamable packages it references sync'ed by default with the CLI
- If a CLI with version A gets invoked in a project that is currently targetting a CLI version B, the CLI will attempt to forward the command to the correct version automatically.
- Current CLI version is now injected into the Dockerfil as an ENV var so that the microservice can be correctly compiled from inside the container.

## Our Dev Workflow Changes
- install.bat/sh scripts are gone now (replaced by parameters on set-package.sh)
- `set-packages.sh` can now optionally re-build only a subset of packages;
- `set-packages.sh` has a parameter now that will cause it to install the CLI built with the local version of the code globally on your machine.
- `sync-rider-run-settings.sh` improved to support parameters on `set-package.sh` passed parameters.


PedroRauizBeamable:
@cdhanna, could use some help to test/update the GHA workflow scripts (as well as testing things locally on MacOS). But... the code here currently works with Docker for Windows (docker commands are running correctly) and UE Editor integration (content and microservices are correctly displayed).

cdhanna (approved):

PR #3500, opened 2024-06-20
adds `beam project enable` command via tags

cdhanna:
Now you can do

```
beam project enable --ids <> --with-group <> --without-group <>
```

The groups correspond to `<BeamServiceGroup>` settings configured in the csproj files for the services. 
The enable command will do the _Set_ math operations.

And then of the services resolved, the command will set the `<BeamEnabled>` flag in the project.


PedroRauizBeamable (approved):
👍

PR #3523, opened 2024-07-05
Fix microservice not booting

gabrielbeamable:
Missing fixes for microservices assembly definition references

### DIFF
--- a/cli/cli/Services/ProjectContextUtil.cs
+++ b/cli/cli/Services/ProjectContextUtil.cs
@@ -560,10 +560,7 @@ public static class ProjectContextUtil
 		const string ITEM_TYPE = "ProjectReference";
 
 		var buildEngine = new ProjectCollection();
-		var stream = File.OpenRead(definition.ProjectPath);
-		var document = XDocument.Load(stream, LoadOptions.PreserveWhitespace);
-		var reader = document.CreateReader(ReaderOptions.None);
-		var buildProject = buildEngine.LoadProject(reader);
+		var buildProject = buildEngine.LoadProject(definition.ProjectPath);
 
 		var references = buildProject.GetItemsIgnoringCondition(ITEM_TYPE).ToArray();
 		for (int i = references.Length - 1; i >= 0; i--)
--- a/cli/cli/Services/ProjectContextUtil.cs
+++ b/cli/cli/Services/ProjectContextUtil.cs
@@ -344,6 +344,7 @@ public static class ProjectContextUtil
 						Path = referencedProject.RelativePath,
 						AssemblyName = referencedProject.BeamUnityAssemblyName
 					});
+					continue;
 				}
 
 				Log.Warning($"Project=[{project.relativePath}] references project=[${referencedProject.FullPath}] but that project was not detected in the beamable folder context. ");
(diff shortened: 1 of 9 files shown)


cli/cli/Services/ProjectContextUtil.cs
```diff
@@ -559,10 +560,7 @@ public static void UpdateUnityProjectReferences(BeamoServiceDefinition definitio
 		const string ITEM_TYPE = "ProjectReference";
 
 		var buildEngine = new ProjectCollection();
-		var stream = File.OpenRead(definition.ProjectPath);
```
gabrielbeamable, line 562: This wasn't working on nightly builds, this particularly line crashed because couldn't parse parts of the csproj that invoke methods to get the actual version. I'm not sure why this was happening, but changing this fixed that and also improved the formatting that was happening in the project when saving it to the file.

client/Packages/com.beamable.server/Editor/UI2/Configs/BeamableMicroservicesSettingsProvider.cs
```diff
@@ -159,10 +161,10 @@ public static SerializedObject GetSerializedSettings(string serviceName)
 			var instance = ScriptableObject.CreateInstance<BeamableMicroservicesSettings>();
 			instance.serviceName = serviceName;
 
-			instance.serviceDefinitions = new List<AssemblyDefinitionAsset>();
+			instance.assemblyReferences = new List<AssemblyDefinitionAsset>();
 			foreach (var name in sd.AssemblyDefinitionsNames)
 			{
-				var guids = AssetDatabase.FindAssets($"{name} t:AssemblyDefinitionAsset");
```
gabrielbeamable, line 165: So, doing `typeof(AssemblyDefinitionAsset).FullName` don't work. The name looks something like `UnityInternals.AssemblyDefinitionAsset`, and if you try to find that in the editor, nothing will pop. So, had to use the name instead of the FullName. Not sure if this fixes the issue you presented in our CR tho, so I'm open to suggestions.

templates/BeamStorage/BeamStorage.csproj
```diff
@@ -15,7 +15,7 @@
         <DotNetConfigPath Condition="'$(DotNetConfigPath)' == '' AND $(DOTNET_RUNNING_IN_CONTAINER)!=true">$([MSBuild]::GetDirectoryNameOfFileAbove("$(MSBuildProjectDirectory)/..", ".config/dotnet-tools.json"))</DotNetConfigPath>
         <DotNetConfig Condition="'$(DotNetConfig)' == '' AND $(DOTNET_RUNNING_IN_CONTAINER)!=true">$([System.IO.File]::ReadAllText("$(DotNetConfigPath)/.config/dotnet-tools.json"))</DotNetConfig>
         <!-- Extracts the version number from the first tool defined in 'dotnet-tools.json' that starts with "beamable". -->
-        <BeamableVersion Condition="'$(BeamableVersion)' == '' AND $(DOTNET_RUNNING_IN_CONTAINER)!=true">$([System.Text.RegularExpressions.Regex]::Match("$(DotNetConfig)", "beamable.*?\"([0-9]+\.[0-9]+\.[0-9]+)\",", RegexOptions.Singleline | RegexOptions.IgnorePatternWhitespace).Groups[1].Value)</BeamableVersion>
+        <BeamableVersion Condition="'$(BeamableVersion)' == '' AND $(DOTNET_RUNNING_IN_CONTAINER)!=true">$([System.Text.RegularExpressions.Regex]::Match("$(DotNetConfig)", "beamable.*?\"([0-9]+\.[0-9]+\.[0-9]+.*?)\",", RegexOptions.Singleline | RegexOptions.IgnorePatternWhitespace).Groups[1].Value)</BeamableVersion>
```
cdhanna, line 18: nice; but I think you need to make this edit in the common template project as well.

cdhanna (changes requested):
just one requirement- please modify the template of the common-lib too.

PR #3539, opened 2024-07-16
fixes #3518 - content ref ids

### LINKED ISSUE #3518: ContentRef should have a constructor that takes a content ID string parameter
I found myself inordinately irritated today that I have to say:
```csharp
var swordRef = new ContentRef<SwordItemContent>
{
  Id = swordContentId
};
```
instead of
```csharp
var swordRef = new ContentRef<SwordItemContent>(swordContentId);
```

Why _can't_ I do the latter? If it is easy to have a string-accepting constructor, we should have one.

cdhanna:
doop- adding lots of constructors.


client/Packages/com.beamable/Common/Runtime/Modules/SharedContent/EventContent.cs
```diff
@@ -42,7 +42,18 @@ public class EventLink : ContentLink<EventContent> { }
...
+		public EventRef(string id) : base(id)
+		{
+			
+		}
+
+		public EventRef()
+		{
+			
+		}
+	} // TODO: Factor
```
gabrielbeamable, line 56: Any idea what this "TODO: Factor" means?

gabrielbeamable (approved):
Just a comment in the code that makes me confused

client/Packages/com.beamable/Common/Runtime/Modules/Shop/ListingRef.cs
```diff
@@ -17,5 +17,15 @@ namespace Beamable.Common.Shop
...
+	public class ListingRef : ContentRef<ListingContent>
+	{
+		public ListingRef()
+		{
+
+		}
+
+		public ListingRef(string id) : base(id)
+		{
+		}
```
allister-beamable, line 29: It is bugging me that there is no blank line in this one like there is in all the rest 🥴
As such, I demand that you keep it in, as a reminder that syntax is independent of whitespace

allister-beamable (approved):
This will make my life a bit easier whenever I am making content refs, so thanks!

How much do we care about linting? Disregarding my jokey comment about the blank line, does it merit a PR-wide ⌘-Alt-L in Rider to keep things consistent for those of us who get a little itchy when things are not in neatly aligned patterns?

PR #3540, opened 2024-07-16
fixes #3395 - date header recog

### LINKED ISSUE #3395: Requester should expose `Date:` header from HTTP response, as a source of server-authoritative time
**Is your feature request related to a problem? Please describe.**

At present if a Game Maker wants to get authoritative server time, they have to implement a C# Microservice method that uses `DateTime.UtcNow` or something equivalent. However, every HTTP response from the Beamable backend includes a `Date:` header that could be used as a source of authoritative server time (assuming the server-to-client latency is small enough to be negligible).

**Describe the solution you'd like**

It would be nice to have a way to directly query `Requester` for the server time reported by the most recent API response. Ideally, this should be anchored to device time in some way. That is, it should allow me to answer two questions:
* "What time was it on the server when that response got sent to me?"
* "How much did device time differ from the server time when we received that response?"

**Describe alternatives you've considered**

While it is possible to work around this issue by implementing my own `[ClientCallable]public Promise<string> GetServerTime()` or the like, that is non-ideal for a couple of reasons: first off it requires me to have a C#MS at all (I might want an otherwise client-oriented game that doesn't use C#MS), and secondly it adds unnecessary network round trips to get information that already exists.

cdhanna:


cdhanna:
> Looks simple and effective. The expectation is that `LatestServerTime` is always coming from the HTTP response headers, and `LatestReceiveTime` is always device-authoritative time, and that they should be within a few milliseconds of each other if the device's clock is set correctly, but diverge a lot if the device's clock is out of synch with the world?

Not quite- If you send a request at t=0 seconds (device time), and you get it back at t=1 second (device time), then the latestReceiveTime will be 1 second, and the latestServer time will be whatever was in the header, which in deviceTime, is probably like .5 seconds or something. Depending on how bad the network latency is, the higher that value will be

allister-beamable:
> Not quite- If you send a request at t=0 seconds (device time), and you get it back at t=1 second (device time), then the latestReceiveTime will be 1 second, and the latestServer time will be whatever was in the header, which in deviceTime, is probably like .5 seconds or something. Depending on how bad the network latency is, the higher that value will be

Now I am confused. Aren't both LatestReceiveTime and LatestServerTime `DateTimeOffset`s? Or are you simplifying the example by saying that t=0 device time is actually 2024-07-16T20:27:07.123Z if the app was started at 20:27:07 today?

If my assumption about the t=0 simplification holds, I take your point about the difference between LatestReceiveTime and LatestServerTime (as also reported by the convenience function `GetLatestServerTimeDifferenceMilliseconds`) being equal to the response latency between server and device. So, revising my expectation: it should normally be a handful to perhaps dozens of milliseconds in most circumstances, plausibly hundreds of milliseconds if there is network congestion, but if it is minutes or hours off then the Game Maker can expect there are device-clock shenanigans going on?

cdhanna:
> > Not quite- If you send a request at t=0 seconds (device time), and you get it back at t=1 second (device time), then the latestReceiveTime will be 1 second, and the latestServer time will be whatever was in the header, which in deviceTime, is probably like .5 seconds or something. Depending on how bad the network latency is, the higher that value will be
> 
> Now I am confused. Aren't both LatestReceiveTime and LatestServerTime `DateTimeOffset`s? Or are you simplifying the example by saying that t=0 device time is actually 2024-07-16T20:27:07.123Z if the app was started at 20:27:07 today?
> 
> If my assumption about the t=0 simplification holds, I take your point about the difference between LatestReceiveTime and LatestServerTime (as also reported by the convenience function `GetLatestServerTimeDifferenceMilliseconds`) being equal to the response latency between server and device. So, revising my expectation: it should normally be a handful to perhaps dozens of milliseconds in most circumstances, plausibly hundreds of milliseconds if there is network congestion, but if it is minutes or hours off then the Game Maker can expect there are device-clock shenanigans going on?

Yes, for me, it was in the hundreds of ms

PR #3548, opened 2024-07-18
[Issue/3480] Services Deploy cancellation

### LINKED ISSUE #3480: CLI - make it possible to cancel the deployment at any moment
Context is that in Unity editor you can cancel the deployment and it makes sense to do that in case you forgot something or decided that you don't want to deploy anymore, however that only closes the window in the Editor, the deployment process continues. Is there already a way to do this? Couldn't find it so leaving this issue here.

gabrielbeamable:
# Ticket

resolves #3480 

# Brief Description

This adds implementation on the ServicesDeployCommand to enable it to be cancelled at any point in time. It also adapts the Unity SDK server implementation to be able to actually cancel the connection between the the SDK and the CLI server.

# Notes


client/Packages/com.beamable/Editor/BeamCli/BeamWebCommand.cs
```diff
@@ -268,6 +274,11 @@ public async Promise Run()
 				//  that is _fine_, but we need to handle it.
 				CliLogger.Log("cli server died, " + ioException.Message);
 			}
+			catch (OperationCanceledException cancelledException)
+			{
+				// A cancellation was requested so the connection was terminated
+				CliLogger.Log("cli command was cancelled, " + cancelledException.Message);
```
cdhanna, line 280: does any extra explicit clean up need to happen with the http object?
```
using HttpResponseMessage response
```

cdhanna (commented):
looks good; just one question about the cleanup

client/Packages/com.beamable/Editor/BeamCli/BeamWebCommand.cs
gabrielbeamable: Nice call! Done!

cdhanna (approved):

PR #3549, opened 2024-07-22
fixes #3339 - removes old static DI ServiceManager

### LINKED ISSUE #3339: Remove old service DI stuff
All the code in `client/Packages/com.beamable/Runtime/Core/Service/` is heckin' old and needs to get deleted.

cdhanna:
This `ServiceManager` was used in DisruptorBeam era days. We left it in for 1.0 as a legacy system. We had a few customers using it, but that user-base doesn't exist anymore.

Best to remove this for the 2.0 release, given we have the opportunity.


Leinnan (approved):
Nice!

PR #3571, opened 2024-08-05
[issue/3510] Handle errors when creating services and storages

### LINKED ISSUE #3510: USAM - editor window not updating correctly after service is has errored
This happens some times while creating a new service, just keeps frozen like in the screenshot and don't update:

Image

gabrielbeamable:
# Ticket

resolves #3510 

# Brief Description

Starts handling errors when creating microservices and storages from USAM. This basically uses the `OnError` event stream of the commands to know that an error happened, and then throw an exception that will be caught by the UI which will update it correctly this time.

# Notes

### DIFF
--- a/client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
+++ b/client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
@@ -499,17 +499,18 @@ namespace Beamable.Server.Editor.Usam
 		/// <param name="name"> The name of the Service/Storage to be created.</param>
 		/// <param name="type"> The type of the Service/Storage to be created.</param>
 		/// <param name="additionalReferences">A list with all references to link to this service.</param>
-		public async Promise CreateService(string name, ServiceType type, List<IBeamoServiceDefinition> additionalReferences)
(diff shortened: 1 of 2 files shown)


client/Packages/com.beamable.server/Editor/Usam/Editor/CodeService.cs
```diff
@@ -870,16 +871,34 @@ private async Promise CreateStorage(string storageName, List<IBeamoServiceDefini
...
+			});
 			await storageCommand.Run();
 
+			await RefreshServices();
+
+			var definition = ServiceDefinitions.FirstOrDefault(def => def.BeamoId == storageName);
+
+			if (definition == null)
+			{
+				await errorPromise;
```
gabrielbeamable, line 897: I'm not really a fan of this promise thing, if something went wrong with the creation but it did create a project anyway, this wouldn't caught that. But the problem with making this better, is that we are starting a new thread for calling the `OnError` stream event, which makes throwing the exception there never getting caught in the creation promise. Is there another way to handle this? I'm not sure if we could just not start new threads for these callbacks, but that would definitely fix this.

PR #3601, opened 2024-08-16
[3600] Fixed Microservice deadlock on windows; added listen capabilities to `fed list` command.

PedroRauizBeamable:
# Ticket

https://disruptorbeam.atlassian.net/browse/BEAM-3600

# Brief Description
fixed issue 3600; refactored Notification server into a couple of utility functions that wrap our websocket auth flow and a listen loop for messages (just so we can make other listen commands)

- Added --listen to `beam fed list` command instead because it makes it so that the data format for both the list and the listen can be the same (this is useful for making the in-engine integrations simpler)

# Notes


cdhanna (approved):
💣 Good show

PR #3604, opened 2024-08-21
Option to get OpenAPI in one document, OpenAPI package update

Leinnan:
# Ticket

Hackaton thing :) 

# Brief Description

> Adds option to combine all OpenAPI documents into one. Useful for auto api generation.

# Checklist

### DIFF
--- a/cli/cli/App.cs
+++ b/cli/cli/App.cs
@@ -28,10 +28,8 @@ using cli.UnrealCommands;
 using cli.Utils;
 using cli.Version;
 using CliWrap;
-using CommandLine.Text;
 using Errata;
 using Microsoft.Build.Locator;
-using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.Extensions.DependencyInjection;
 using Newtonsoft.Json;
 using Serilog;
@@ -39,7 +37,6 @@ using Serilog.Core;
 using Serilog.Enrichers.Sensitive;
 using Serilog.Events;
 using Spectre.Console;
-using Spectre.Console.Rendering;
 using System.CommandLine;
 using System.CommandLine.Builder;
 using System.CommandLine.Help;
@@ -176,7 +173,7 @@ public class App
 		_serviceConfigurator?.Invoke(services);
 	}
 
-	private Action<IDependencyProvider> setLogger = _ =>
+	private Action<IDependencyProvider> _setLogger = _ =>
 	{
 		// no-op
 	};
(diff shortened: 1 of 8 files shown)


cli/cli/Services/SwaggerService.cs
```diff
@@ -419,10 +415,126 @@ void RewireSchema(OpenApiSchema schema)
...
 	public async Task<IEnumerable<OpenApiDocumentResult>> DownloadBeamableApis(BeamableApiFilter filter)
 	{
-		var selected = Apis.Where(filter.Accepts).ToList();
+		var selected = APIS.Where(filter.Accepts).ToList();
 		return await DownloadOpenApis(_downloader, selected).ToPromise(); //.ShowLoading("fetching swagger docs...");
 	}
 
+	public OpenApiDocument GetCombinedDocument(IEnumerable<OpenApiDocumentResult> apis)
+	{
+		var combinedDocument = new OpenApiDocument(apis.FirstOrDefault().Document);
```
cdhanna, line 424: in general, do we need to be worried about duplicates? 
_if_ there was a duplicate, I think the `Add` function would throw an exception right? So I guess in that sense, we are covered?

Leinnan: And how should we approach them here?

cli/cli/App.cs
```diff
@@ -176,7 +173,7 @@ private void ConfigureServices(IDependencyBuilder services)
 		_serviceConfigurator?.Invoke(services);
 	}
 
-	private Action<IDependencyProvider> setLogger = _ =>
+	private Action<IDependencyProvider> _setLogger = _ =>
```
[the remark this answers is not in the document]
Leinnan: Reducing warnings, I can reverse that.

cli/cli/Services/SwaggerService.cs
Leinnan: Update - so I dig more into our `SwaggerService` code, and now it resolves duplicates based on the code that we already use for Unity/Unreal code generation. There are 3 approaches:
- if there is `OpenApiPathItem` duplicate, I just skip it, should I care?
- if there is a `OpenApiSchema` or `OpenApiSecuritySchema` duplicate I check if it is exactly the same as existing one, if yes then skip, if no it will try to add it and throw Exception
- if there is other type of duplicate it will try to add it and throw Exception, there is none ATM

cdhanna: I think all that logic is fine :+1:

cdhanna (approved):

PR #3607, opened 2024-08-23
[3606] `project ps -w` refactoring

PedroRauizBeamable:
# Ticket

https://disruptorbeam.atlassian.net/browse/BEAM-XXXX

# Brief Description

- Replaced NetMQ service discovery with a simple socket broadcast/listen.
- Adjusted `CheckStatusCommand` to provide events about all known services when invoked before beginning to listen
- Adjusted `CheckStatusCommand` to provide routing key information whenever it is available
- OpenSwaggerCommand now takes `--routing-key` as an option as opposed to `--remote`. When no routing-key is given, it'll talk to the remote running service.

# Checklist


Leinnan:
I like it, one thing that bothers me- isn't it a breaking change because of the flag thing?

PedroRauizBeamable:
@Leinnan I don't _think_ this is a relevant breaking change, though. AFAIK Unity doesn't use that command yet (only 2.0 will use) and Unreal's usage of it will be fixed with 1.0.0 which is dependent on CLI 2.1.0 when this goes out. If you feel really strongly about it, I guess we can:

- Add it back along with a --local flag
- Make sure they are mutually exclusive (only one of the three flags are allowed)
- Map `--remote` to empty `routing key`
- Map `--local` to the locally generated `routing key`

@cdhanna Any thoughts on the above ☝️ ?

Also, found a couple of bugs on other commands that relied on the `project ps` not emitting the initial "one event per existing service". I'll be pushing the fixes here in the afternoon.

cdhanna (approved):
met in Discord and 👍

PR #3611, opened 2024-08-26
converge on cli.sln as the only dev sln

cdhanna:
We've had two main development `.sln` files, the `microservice.sln` and `cli.sln`. 

This PR takes the (few) projects that were unique to `microservice.sln` and adds them to `cli.sln`. Now, as a developer, we can operate out of the `cli.sln` as the "base beamable code" solution; and as we are testing an individual  SAM, that sln as well. 

I didn't bother "fixing" the stale readme, but I deleted the stuff that was blatantly _wrong_, and I removed the `microservice.sln` completely.


PedroRauizBeamable (approved):
Sooo much value in pressing delete.

Issue #3686, opened 2024-09-25
CLI- group commands


Leinnan:
Add commands:
- `beam project group` lists groups in projects
- `beam project group add <SERVICE_NAME> <GROUP_NAME>` adds group to project
- `beam project group rm<SERVICE_NAME> <GROUP_NAME>` remove group from project

Also add new `--group <GROUP_NAME> <GROUP_NAME>` to `beam project new` commands to add groups during services creation.

PR #3687, opened 2024-09-25
Beam groups commands

Leinnan:
Implements #3686

### DIFF
--- a/cli/cli/Commands/Project/GroupCommand.cs
+++ b/cli/cli/Commands/Project/GroupCommand.cs
@@ -1,0 +1,43 @@
+using Beamable.Common.Semantics;
+using Spectre.Console;
+
+namespace cli.Commands.Project;
+
+public class GroupCommand : CommandGroup
+{
+	public GroupCommand() : base("group", "Generate an ignore file in .beamable folder for given VCS")
+	{
+	}
+
+	public override void Configure()
+	{
+	}
+
+	public override Task Handle(CommandGroupArgs args)
+	{
+		var table = new Table();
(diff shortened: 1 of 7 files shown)


PedroRauizBeamable (changes requested):
Seems fine-ish. Couple of things:
- The add/remove commands should take multiple groups. 
- The code that operates on the system data, should be on the system. Not in the `Definition` class. 
- The `SetBeamGroup` is not very flexible --- See recommendation in comment.

cli/cli/Commands/Project/GroupCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+		{
+			AnsiConsole.WriteLine("No services groups found");
+		}
+
+		return Task.CompletedTask;
+	}
+}
+
+
+public class UpdateGroupArgs : CommandArgs
```
PedroRauizBeamable, line 45: In general, try to avoid using API-facing types (`CommandArgs`) as the same type in the internal system's API there. Interface types have all sorts of backward compatibility concerns that internal systems don't. Its less likely for people to accidentally break things if these are separate types (even if their schemas are mostly the same, semantically they mean very different things --- the API one means "this is how users interact with the CLI" and the internal system means "these are the ways we can modify groups all through our code-base").

PedroRauizBeamable (approved):
Left a small coment there --- there's not a lot of harm in this particular case to leave it as is, but... Still a good thing to think about "what the type means" in the code before reusing it just due to schema similarities.

PR #3699, opened 2024-10-09
[3696] Added new BeamableSourceGenerator project and federation declaration workflow

PedroRauizBeamable:
# Ticket
https://github.com/beamable/BeamableProduct/issues/3696

# Brief Description

- Reorganized Federation files;
- IThirdPartyCloudIdentity is now Obsolete (replaced with IFederationId) --- functionally the same.
- Added `beam fed add/remove/list` commands to interact with the `BeamSourceGenConfig.json` file of each Microservice.
- `beam project ps` now correctly lists federations for each routing key pulling from the correct source based on whether the service is running locally or remotely.
- Moved some DLL loading code out of the `beam project generate-client` command (just some book-keeping refactoring)
- Fixed issue in UE code-generation that would cause some Obsolete warnings on certain JSON serialization output code.
- `BeamoLocalManifest` initialization now ensures the existance of `BeamSourceGenConfig.json` for all microservices.
- Finding federation components for OAPI and RouteBuilding stuff now looks for the parent `IFederation` interface instead of a list of federation names (eventually, we might not need to have these things be reflection based and we can move them into the source generator).
- Added an `AdditionalFiles` entry to the `Beamable.Microservice.Runtime.props` that ensures the `BeamSourceGenConfig.json` file is visible to our source generator.
- `BeamService` template now includes a reference to our `Beamable.Microservice.SourceGen` as an Analyzer.
- Modified Unity-specific code to handle deprecation of `IThirdPartyCloudIdentity`

Internal Project Structure Changes:
- Added `.beamable` folder to `templates` directory so that the templates for microservices actually function properly (makes testing and working in the `templates` solution easier).
- Added `ProjectReferences` to the `templates` solution to make it easier to develop the source generator (TODO: we should really merge all these projects in a single solution).
- Added glob pattern to remove all bin/obj folders that are kept inside /cli and /template directories
- Added the Source Generator project to the `set-packages.sh` script.

# Notes


PedroRauizBeamable:
Closed because these changes went into main bundled up in another PR.

PR #3703, opened 2024-10-11
add json string

cdhanna:
Adds a new `JsonString` type to the `JsonSerializable` namespace that has magical properties. If you say its name three times in a mirror, Jason will stringify into existence and turn your bits into bytes and take you to dream town. 

But also, the class will act as a "whole-sale" json implacement operator. When serializing, it writes JSON raw into the stream. When deserializing, we already have a parsed value, so hopefully that is good enough... but if it isn't, we can serialize the `ArrayDict` into more json, and then deserialize that json into a hard type. 

The openAPI takes the field with the json extension, and converts a version of this type.


cli/beamable.common/Runtime/JsonSerializable/JsonSaveStream.cs
```diff
@@ -278,6 +278,17 @@ protected void AppendEscapedCharacter(char c)
 			}
 
 
+			public bool SerializeNestedJson(string key, ref JsonString jsonString)
+			{
+				// don't serialize the field at all if the json is null.
+				if (jsonString == null) return true;
```
gabrielbeamable, line 284: Shouldn't be false here instead, since it couldn't serialize the thing?

cli/beamable.common/Runtime/JsonSerializable/LoadStream.cs
```diff
@@ -70,6 +71,15 @@ public void DeregisterISerializableFactory(ISerializableFactory factory)
 				factories.Remove(factory);
 			}
 
+			public bool SerializeNestedJson(string key, ref JsonString jsonString)
+			{
+				if (curDict.TryGetValue(key, out var tmp))
+				{
+					jsonString = JsonString.FromValue(tmp);
+				}
+				return true;
```
gabrielbeamable, line 80: Same question as above here, what is the purpose of the bool return if it will always be the same?

gabrielbeamable (approved):

cli/beamable.common/Runtime/JsonSerializable/JsonSaveStream.cs
PedroRauizBeamable: I _think_ it's that, in these cases, the (de)serialization is always valid. As in, if the JSON is `null`, the correct serialization is not writing anything so... we just don't write anything and return `true` to say "this is fine".

cli/beamable.common/Runtime/JsonSerializable/LoadStream.cs
PedroRauizBeamable: Same idea here --- when deserializing, if we have some value in the JSON blob, we create a `JsonString` from it --- otherwise, we don't create the `JsonString` instance.

PedroRauizBeamable (approved):
Looks good to me

PR #3710, opened 2024-10-17
fixes window dotnet install

cdhanna:
the windows version of the install string was using `Channel` instead of `Version`. Oops.
image

Also, now that we execute `beam` as `dotnet beam`; we don't need to append the `.exe` for windows.


Leinnan (approved):
I still have issues with CLI, but dotnet did install correctly and it is no blocked on it! 🔥 
image

PR #3775, opened 2024-11-11
Feature/lootbox lightbeam

gabrielbeamable:
This PR makes the following:
 - Adds a new lightbeam with a microservice
 - Adds a filter to ignore certain folders when loading projects in the CLI


client/Packages/com.beamable/Samples/LightBeamSamples/LootBox/BeamableServices~/LootBoxService/LootBoxService.cs
```diff
@@ -0,0 +1,46 @@
+using Beamable.Common;
+using Beamable.Server;
+using System;
+using System.Globalization;
+
+namespace Beamable.LootBoxService
+{
+	[Microservice("LootBoxService")]
```
gabrielbeamable, line 8: Create a sibling folder of `BeamableServices~` with an assembly definition inside called `<beamoId>.Client` and make sure there is a empty script in there

PR #3789, opened 2024-11-22
switches federation to attr

cdhanna:
Instead of doing 
```csharp
public class Tuna : IFederationId {
 public string UniqueName => "lizard attack!"
}
```

now you do
```csharp
[FederationId("lizard attack!")]
public class Tuna : IFederationId {}
```

The advantage is that our source generator can _read_ this information when it actually needs to. So put differently, this approach works; where as the old one didn't for Unity. 

However, we should keep the interface constraint, because it allows the type itself to be used in constrained generics. 
I've added a `GetUniqueName()` _method_ to backfill; although since extension props are not supported, it will require user intervention. 

I also removed the code-gen part of the code-gen; so that developers need to manually type in the federation themselves, always.


gabrielbeamable (changes requested):
I think it's missing a change to the `FederationId.cs.txt` template file, so when we create a federation id through the Unity editor, it will create with the attribute and not the property. The file is in `Packages/com.beamable.server/Editor/UI2/ScriptTemplates/FederationId.cs.txt`

gabrielbeamable (approved):

PR #3813, opened 2024-12-10
Unity 6 upgrade

gabrielbeamable:
Upgrading the Unity SDK to work with all three versions 2021, 2022 and Unity 6.

Most of the fixes are in changing deprecated UI toolkit stuff.


cdhanna (commented):
this looks pretty good! I'm happy that we are moving away from UIToolkit in general so that these #if/def style things are more palatable. 

Once the CI jobs are passing, I think we can give this the ol' mergo

PR #3835, opened 2024-12-19
[BEAM-3628] Feature/sim type editor

### LINKED ISSUE #3628: Update SimGameType's Federation to use the nice UX dropdown
https://github.com/beamable/BeamableProduct/blob/1.19.22/client/Packages/com.beamable/Common/Runtime/Modules/SharedContent/SimGameType.cs

The `ItemContent` uses a federation with a nice dropdown of available C#MS options; we should apply that to the simGameType as well.

gabrielbeamable:
# Ticket

resolves #3628 

# Brief Description

Added custom property drawer to federations in `SimGameType` scriptable objects

### DIFF
--- a/cli/beamable.common/Runtime/Modules/SharedContent/SimGameType.cs
+++ b/cli/beamable.common/Runtime/Modules/SharedContent/SimGameType.cs
@@ -67,7 +67,7 @@ namespace Beamable.Common.Content
 	[ContentType("game_types")]
 	[Serializable]
 	[Agnostic]
-	public class SimGameType : ContentObject
+	public class SimGameType : ContentObject, ISerializationCallbackReceiver
 	{
 		[Obsolete("Use `CalculateMaxPlayers` instead.")]
 		[IgnoreContentField]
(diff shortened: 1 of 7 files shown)


cli/beamable.common/Runtime/Modules/SharedContent/SimGameType.cs
```diff
@@ -140,6 +153,14 @@ public void OnAfterDeserialize()
 				};
 			}
 
+			if (federatedGameServerNamespace.HasValue && !newFederatedGameServerNamespace.HasValue && isFirstTime)
+			{
+				isFirstTime = false;
+				newFederatedGameServerNamespace.HasValue = true;
+				var nameSpace = new Namespace {Name = federatedGameServerNamespace};
+				newFederatedGameServerNamespace.Value = nameSpace;
```
Leinnan, line 161: You would need to check but if I remember correctly there are implicit operator methods for that so you should be able to just do 
```suggestion
				newFederatedGameServerNamespace = new Namespace {Name = federatedGameServerNamespace};
```

cdhanna (approved):

PR #3847, opened 2025-01-14
fixes #3839 - adds pid based self-destruct to microservices

### LINKED ISSUE #3839: CLI / SAMS - watch process-id for auto-exit
The CLI `beam project run` command should accept a new option flag, `--watch-process-id`. The CLI should forward this to a new setting in SAMS, `BEAM_WATCH_PROCESS_ID` that when present, spins up a daemon process that constantly checks for the system's procs, and if the watched process id is absent, then the SAM should insta-die. 

This is useful because it removes the dangling processes when you shutdown UE or Unity.

cdhanna:
There is a new ENV var in the C#MS framework that when present locally, kicks off a monitor loop checking for the process existence- when the process no longer exists, the microservice terminates. 

You'll see logs like this,
```
[Info] Quitting because required process no longer exists

[Debug] Shutdown started... 0 tasks running.

[Debug] sending request 
{"id":-8,"method":"delete","path":"gateway/provider","from":null,"body":{"type":"basic","name":"micro_Fishsticks","beamoName":null,"routingKey":null,"startedById":null}
}
```

The CLI sets this ENV var based on the existence of a new option flag.

### DIFF
--- a/cli/beamable.common/Runtime/Constants/Constants.cs
+++ b/cli/beamable.common/Runtime/Constants/Constants.cs
@@ -164,6 +164,7 @@
 			public const string BEAM_PATH = "BEAM_PATH";
 			public const string BEAM_DOTNET_PATH = "BEAM_DOTNET_PATH";
 			public const string BEAM_DOTNET_MSBUILD_PATH = "BEAM_DOTNET_MSBUILD_PATH";
+			public const string BEAM_REQUIRE_PROCESS_ID = "BEAM_REQUIRE_PROCESS_ID";
 		}
 	}
 }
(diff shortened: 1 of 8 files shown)


PedroRauizBeamable (approved):

gabrielbeamable (approved):
Are we going to use this on Unity SDK to auto terminate the microservice when the Unity editor is closed for example?

cdhanna:
> Are we going to use this on Unity SDK to auto terminate the microservice when the Unity editor is closed for example?

Yes, exactly. Need to make a ticket to do that part of this.

PR #3849, opened 2025-01-15
player init federation

cdhanna:
This PR also includes a bunch of auto-gen stuff. 
Mostly, it is actually a very small amount of code to support the new flow.


cli/beamable.common/Runtime/Federation/FederatedPlayerInit.cs
```diff
@@ -0,0 +1,42 @@
...
+		Promise CreatePlayer(PlayerInitiation player);
+	}
+
+	/// <summary>
+	/// The <see cref="PlayerInitiation"/> data contains the <see cref="Account"/> information
+	/// for the player that is initialization, as well as the initProperties provided in the
+	/// token creation request. 
+	/// </summary>
+	[Serializable]
+	public class PlayerInitiation
```
cdhanna, line 29: @jstnlef can this type appear from the auto-gen?

microservice/beamable.tooling.common/Microservice/CallableAdapter.cs
```diff
@@ -83,6 +83,66 @@ public List<ServiceMethod> ScanType(MicroserviceAttribute serviceAttribute, Serv
...
+			{
+				var err = $"Unable to map method name to path part. name=[{method.Name}]";
+				throw new Exception(err);
+			}
+			
+			// we use "default" as a hardcoded value here because this particular federation ignores
+			//  the registered namespace and uses the term, "default" always anyway. 
+			//  this means that the user can pick whatever string they want, and it literally has no
+			//  effect. :shrug: 
+			var path = $"default/{pathName}";
```
cdhanna, line 125: put this back to namespace-
And add a thing to the source generator that checks that if the user has a playerInit federation, the federation id MUST be "default"

PR #3855, opened 2025-01-20
[BEAM-3843] do best try finding asmdefs

### LINKED ISSUE #3843: 2.0.2 update flow issue
We received a bug report with this null reference will updating from 1.19.23 to 2.0.2

```
System.NullReferenceException: Object reference not set to an instance of an object
  at Beamable.Server.Editor.Usam.UsamMigrator.CreatePlan (Beamable.Server.Editor.MicroserviceReflectionCache+Registry registry, Beamable.Server.Editor.Usam.UsamService usam) [0x006db] in .\Library\PackageCache\com.beamable.server@2.0.2\Editor\Usam\Editor\UsamMigrator.cs:670 
  at Beamable.Server.Editor.Usam.UsamService.LoadLegacyServices () [0x00001] in .\Library\PackageCache\com.beamable.server@2.0.2\Editor\Usam\Editor\UsamService.cs:594 
```

gabrielbeamable:
# Ticket

resolves #3843 

# Brief Description

This First add fallback for posterior errors and doc and then it does a best try finding asmdefs by their property names as well.


client/Packages/com.beamable.server/Editor/Usam/Editor/UsamMigrator.cs
```diff
@@ -752,12 +755,27 @@ public static List<AssemblyDefinitionAsset> GetAssemblyDefinitionAssets(IDescrip
...
+								}
+							}
 						}
+						break;
+					}
 
-						asset = AssetDatabase.LoadAssetAtPath<AssemblyDefinitionAsset>(assetPath);
+					if (asset == null) //if the asset is still null, we don't try to add this assembly as reference, put it in a list and ask user to manually add the reference
+					{
+						errors.Add($"The assembly reference {name} could not be added, please manually add it after the migration is completed.");
```
Leinnan, line 777: Here I would add more info- descriptor is missing that reference. If user has multiple services it can be useful to have that specific info.

Leinnan: Oh yeah, you are right. My main point was is to specify which service is affected so if I have info about adding it manually after migration I know `which` service is missing what.

client/Packages/com.beamable.server/Editor/Usam/Editor/UsamService_Reload.cs
```diff
@@ -21,6 +21,7 @@ public Promise WaitReload()
 			var taskId = ++latestReloadTaskId;
 			
 			LoadLegacyServices();
+			LoadAllAssemblies();
```
cdhanna, line 24: is it possible to get this to work without calling `LoadAllAssemblies` in Reload()? I know `Reload()` gets called multiple times during a domain reload, right? 
Or rather; maybe in the `LoadAllAssemblies()` function, only call `EnumerateAssemblyDefinitionAssets` when `allAssemblyAssets` is null?

client/Packages/com.beamable.server/Editor/Usam/Editor/AssemblyUtil.cs
```diff
@@ -98,6 +100,32 @@ private void UpdateReferencedAssemblies(List<BeamUnityAssemblyReferenceData> ref
 			}
 		}
 
+		public string GetAssemblyName(BeamUnityAssemblyReferenceData assemblyData)
+		{
+			var asset = _usam.allAssemblyAssets.FirstOrDefault(a => a.name.Equals(assemblyData.AssemblyName));
```
cdhanna, line 105: is it possible to pre-compute these names with a Dictionary instead of needing to scan through all the assemblies every time? 
When you load up all the assemblies, I think you could do a single pass, and create a dictionary that maps from assembly name to asset, and vice-verca?

cdhanna (changes requested):
yahoo! the root of my PR is about the performance of this change. As much as possible, I'd like to precompute all the assembly data, and then access it through maps and such when you're scanning through the deps in the migrator step.

client/Packages/com.beamable.server/Editor/Usam/Editor/UsamService_Reload.cs
gabrielbeamable: I'll test to see if it's being called multiple times, didn't notice that before

PR #3856, opened 2025-01-21
[BEAM-3585] Add troubleshoot for deployment error

### LINKED ISSUE #3585: USAM - deploy sometimes break before uploading images
In the deployment process, first we build all images and deploy to local, which means that we run those services and storages that are being deployed locally using docker. After that, we check if those services are indeed running by making a request to the `/Docs` endpoint of those services. Sometimes, for some unknown reason, those services are exiting before that request is being made, which causes the backend to return a 500 error like this:

```
CLI: stderr,[[/basic/1709927373427153.DE_1709927373427156.MacBook-Promicro_JustAnotherOne/admin/Docs]]request error with response code: 500 and message: No binding found for service 1709927373427153.de_1709927373427156.macbook-promicro_justanotherone.basic with routingKeys: ,
```

However, if you look into the logs of the service that exited, you will see this error:

```
2024-08-12 17:17:27 assertion failed [block != nullptr]: BasicBlock requested for unrecognized address
2024-08-12 17:17:27 (BuilderBase.h:550 block_for_offset)
```

As mentioned before this is not consistent, most of the times it works fine.

gabrielbeamable:
# Ticket

resolves #3585 

# Brief Description

Adds documentation for a known issue with deployment in M1, Mx computers.

### DIFF
--- a/cli/cli/Docs/Guides/5_Microservice-Deployment.md
+++ b/cli/cli/Docs/Guides/5_Microservice-Deployment.md
@@ -260,7 +260,7 @@ When you do that, the service will not start --- but the container will look exa
 
 > 📘 Troubleshoot
 > 
-> If you are on a mac with the apple silicon processors (M1, M2, etc) the following error might occur when deploying C# Microservices. In that case, make sure that the `Use Rosetta for emulation on Apple Silicon`is disable in your Docker settings.
+> If you are on a Mac with the apple silicon processors (M1, M2, etc) the following error might occur when deploying C# Microservices. In that case, make sure that the `Use Rosetta for emulation on Apple Silicon` is disabled in your Docker settings.
 > 
 > ```
 > assertion failed [block != nullptr]: BasicBlock requested for unrecognized address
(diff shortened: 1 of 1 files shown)


cli/cli/Docs/Guides/5_Microservice-Deployment.md
```diff
@@ -257,3 +257,12 @@ ENTRYPOINT ["tail", "-f", "/dev/null"]
 ```
 
 When you do that, the service will not start --- but the container will look exactly like it does just before the service runs, except it won't be cleaned up by the health-check failing. This means you can easily inspect its file structure via Docker for Windows/Mac's UI or other tools.
+
+> 📘 Troubleshoot
+> 
+> If you are on a mac with the apple silicon processors (M1, M2, etc) the following error might occur when deploying C# Microservices. In that case, make sure that the `Use Rosetta for emulation on Apple Silicon`is disable in your Docker settings.
```
allister-beamable, line 263:
```suggestion
> If you are on a Mac with the apple silicon processors (M1, M2, etc) the following error might occur when deploying C# Microservices. In that case, make sure that the `Use Rosetta for emulation on Apple Silicon` is disabled in your Docker settings.
```

PR #3862, opened 2025-01-23
Lobby leave docs update

Leinnan:
Docs update

### DIFF
--- a/client/Packages/com.beamable/Runtime/Player/PlayerLobby.cs
+++ b/client/Packages/com.beamable/Runtime/Player/PlayerLobby.cs
@@ -249,6 +249,9 @@ namespace Beamable.Player
 		/// <summary>
 		/// Leave the lobby if the player is in a lobby.
 		/// </summary>
+		/// <remarks>The implemented behavior is that when the lobby host leaves, the lobby is destroyed.
+		/// If there’s a need for the host to leave the lobby without destroying it,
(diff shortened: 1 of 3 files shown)


client/Packages/com.beamable/Runtime/Player/PlayerLobby.cs
```diff
@@ -249,6 +249,9 @@ public async Promise KickPlayer(string playerId)
 		/// <summary>
 		/// Leave the lobby if the player is in a lobby.
 		/// </summary>
+		/// <remarks>The implemented behavior is that when the lobby host leaves, the lobby is destroyed.
+		/// If there’s a need for the host to leave the lobby without destroying it,
+		/// the host should first appoint a new lobby host using the Update method.</remarks>
```
cdhanna, line 254: careful, looks like this comment is only on the /Packages based path... but this file is auto-copied from the corresponding one in beamable.common/.

cdhanna (changes requested):

client/Packages/com.beamable/Runtime/Player/PlayerLobby.cs
Leinnan: Wait, isn't `PlayerLobby` class Unity only? It is not the  _clean_ API one.

cdhanna (approved):
my bad; not an auto-gen file :+1:

PR #3871, opened 2025-01-30
New method for switching realms in runtime

Leinnan:
One of our customers ended up in endless loading loop because of the fact that `Beam.ChangePid` method is not checking if the realm they wanted switch to isn't already active. I was thinking about that case and I think this new method would be a improved way to handle that case. I didn't want to change return type in the old one so it would not end up being breaking change.


cdhanna (approved):
nice

PR #3872, opened 2025-01-31
fixed #3865 config window gui exceptions

### LINKED ISSUE #3865: [Unity] Enter playmode while in the Config tab of BeamableServices breaks
If you are in the `Config` tab of the `Beamable Services` window in Unity, and you enter play mode, then a bunch of errors happen, as showed below:

```
GUI Error: Invalid GUILayout state in  view. Verify that all layout Begin/End calls match
UnityEngine.GUIUtility:ProcessEvent (int,intptr,bool&) (at /Users/bokken/build/output/unity/unity/Modules/IMGUI/GUIUtility.cs:189)
```

Image

cdhanna:
two real issues here, 
1. we had an early `return` in one of the UI helper methods that wasn't calling the required `EndVertical()` function, so that was the cause of the error.
2. the _intent_ of the UI is to flag the fact that the serialized objects are broken; so I added some code earlier up that will check if the objects are null, and if so, re-initialize them.


gabrielbeamable (approved):
nice!

PR #3876, opened 2025-02-06
[BEAM-3830] require user to install dotnet

### LINKED ISSUE #3830: [Unity SDK 2.1.0] SDK gets stuck after installing on Unity 6
## Describe the Bug 
Toolbox button is nowhere to be found after installation

Workaround:
Relaunch Editor

1. Launch Unity Editor 6
2. Install SDK 2.1.0 RC2

### Expected Behavior
Toolbox button must appear after installation

### Actual Behavior
Toolbox button is missing

## Screenshots
Toolbox_button

## Metadata
2.1.0 RC2
Unity 6000.0.31f1

gabrielbeamable:
# Ticket

resolves #3830 

# Brief Description

Instead of attempting a local installation of dotnet, prompt the user to download and install the dotnet of version `8.0.302` from dotnet's github. This also changes how we check if dotnet is installed or not, by running the `dotnet --info` command.


client/Packages/com.beamable/Editor/Dotnet/DotnetUtil.cs
```diff
@@ -113,38 +89,75 @@ public static bool InstallLocalManifest()
 
 		static void InstallDotnetToLibrary()
 		{
-			EditorUtility.DisplayProgressBar("Downloading Dotnet", "Beamable is fetching dotnet installer", .1f);
-			DownloadInstallScript();
+			if (EditorUtility.DisplayDialog("Dotnet Installation Required", "Beamable Unity SDK requires dotnet sdk 8.0.302 to function properly. Please download the  SDK Installer and proceed with the installation before continuing.","Download", "Close"))
+        	{
+				Application.OpenURL("https://github.com/dotnet/core/blob/main/release-notes/8.0/8.0.6/8.0.302.md");
```
gabrielbeamable, line 94: We should direct this to the download link directly instead of to the page with all the links, use the same code that we do for the Docker installation:
`GetDockerDownloadLink()` has this working for docker installation, but it runs in CLI, should test if it works on windows as well.

client/Packages/com.beamable/Editor/Dotnet/DotnetUtil.cs
```diff
@@ -113,38 +89,75 @@ public static bool InstallLocalManifest()
 
 		static void InstallDotnetToLibrary()
 		{
-			EditorUtility.DisplayProgressBar("Downloading Dotnet", "Beamable is fetching dotnet installer", .1f);
-			DownloadInstallScript();
+			if (EditorUtility.DisplayDialog("Dotnet Installation Required", "Beamable Unity SDK requires dotnet sdk 8.0.302 to function properly. Please download the  SDK Installer and proceed with the installation before continuing.","Download", "Close"))
```
gabrielbeamable, line 92: If they hit `Download` popups another dialog different from this one that says "Waiting for dotnet installation before proceeding" with a `Ok`

client/Packages/com.beamable/Editor/Dotnet/DotnetUtil.cs
```diff
@@ -113,38 +89,75 @@ public static bool InstallLocalManifest()
...
+				CreateNoWindow = true,
+				RedirectStandardOutput = true,
+				RedirectStandardError = true,
+			};
+			proc.Start();
+			proc.WaitForExit();
+
+			var output = proc.StandardOutput.ReadToEnd();
+
+			string sdkSection = output.Split(new string[] {".NET runtimes installed:"}, StringSplitOptions.None)[0];
```
gabrielbeamable, line 119: we should force English as the language for the command, in case the machine is configured with a different language. SImilar to this:
```csharp
public static Command GetDotnetCommand(string dotnetPath, string arguments)
	{
		var command = Cli.Wrap(dotnetPath)
			.WithEnvironmentVariables(new Dictionary<string, string> { ["DOTNET_CLI_UI_LANGUAGE"] = "en" })
			.WithArguments(arguments);
		return command;
	}
```

PR #3889, opened 2025-02-14
fixes #3888

### LINKED ISSUE #3888: SAMs Content cannot load from other assembly
loading Content types from other assemblies was not working. Those assemblies were _referenced_, but not _loaded_ when the Beamable content sweep happened.

cdhanna:
manually load all referenced assemblies during Microservice bootup.


PedroRauizBeamable (approved):
I'm curious about the perf implications of this in Unity (if it makes any noticeable difference in the domain reload time/relative to when we don't do this load)

gabrielbeamable (approved):

Issue #3893, opened 2025-02-18
[Unity SDK 2.1.0] Search textbox disappears in Beam Services when there are a lot of logs


BeamDwayne:
## Describe the Bug 

Service logs makes Search textbox disappear if there are 300+ logs

### Steps to reproduce the behavior:

1. Launch Unity Editor
2. Install Beamable SDK 2.1.0 rc3
3. Login/Register
4. Launch Docker
5. Open Beam Services
6. Create a Service
7. Play Service

### Expected Behavior

Search textbox wont disappear

### Actual Behavior

Search textbox disappear if the logs are 300+

## Screenshots

![Image](https://github.com/user-attachments/assets/ac8a4ade-4a33-42e2-be43-985da166572e)

## Metadata

- SDK version: 2.1.0_RC3
- Engine version: Unity 6
- Operating System: Windows 10

## Additional Context
Unable to reproduce in storage since it only shows few logs

cdhanna:
This is actually caused by the window width. 

![Image](https://github.com/user-attachments/assets/b3cda4b0-f349-48a3-b940-32713c7d660f)

BeamDwayne:
Thank you for this Chris, but is it still a valid issue or can I close it?

cdhanna:
to get started, look at `client/Packages/com.beamable/Editor/BeamCli/UI/LogUtil/LogViewUtil.cs` in the `DrawLogWindow` function.

cdhanna:
While fixing this, we should also figure out how to make the search bar take the full available width. 

<img width="461" alt="Image" src="https://github.com/user-attachments/assets/3ac288d4-3640-47c6-8b77-7b6746b11f73" />

PR #3928, opened 2025-03-07
Fix issue which SearchBar wasn't being rendered when Window doesn't have enough width [3893]

DiasAtBeamable:
# Ticket

https://github.com/beamable/BeamableProduct/issues/3893

# Brief Description

> It fixes the issue of there beingn't enough space for SearchBar to render.
> Adds a button when there isn't enough space for the SearchBar to notify the user to resize the window.

https://github.com/user-attachments/assets/81a99fca-e00b-4148-9a19-3fdb027ab7f0

# Checklist


cdhanna (approved):
💣

PR #3935, opened 2025-03-11
Update the `Select Realm` menu to allow updating config-defaults

Leinnan:
Improve the workflow and provide some explanation why config-defaults exists and what it does.
Zrzut ekranu 2025-03-11 o 15 14 50


cdhanna:
I like the `save to config-defaults` button; but the UX around the text information doesn't feel idiomatic.

Leinnan:
> I like the `save to config-defaults` button; but the UX around the text information doesn't feel idiomatic.

It was looking as good idea at first, but yeah. I've removed it.

cdhanna (approved):

allister-beamable (approved):
I forgot to review this earlier, but I like it! The `[build]` suffix will remind people that there is a build realm, and the presence of a "Save to config-defaults" item will help reinforce the concept when it differs from the one selected in editor. Ship it!

PR #3937, opened 2025-03-12
Open portal on user realm

### LINKED ISSUE #3936: Open Portal command doesn't use the customer realm
## Describe the Bug 

Open Portal command does not respect the customer realm, so if user context is on different realm than the one selected in editor it still uses the one selected in editor.

Run the game.
On start of the game SwitchPid to different one than the one in editor.
Open Admin Console.
Run 'portal' command.

### Expected Behavior

Opens beamable portal on the realm of the user.

### Actual Behavior

Opens beamable portal on the realm of the editor.

Leinnan:
Open Portal command does not respect the customer realm, so if user context is on different realm than the one selected in editor it still uses the one selected in editor.

Fixes #3936


cdhanna (approved):
looks good- but please add a changelog line

PR #3939, opened 2025-03-13
Fix/switch pid validation

### LINKED ISSUE #3938: Switch Pid should validate the string input
Recently on the call with customer we've spend way too much time because of the fact that the Pid input for the `Beam.ChangePid` method was missing `DE_` in the beginning. I think we could and should validate whether the input was valid and in case it was not provide a log message with information about invalid input and what input would be valid(at least in the editor we have a list of the valid realm ids).

Leinnan:
Fixes #3938


cdhanna (approved):
Its fine.
I'd be a noodge and suggest you don't use regex for something so simple... 
Its, the first char must be `D`, the second char must be `E`, the third char must be `_`, all other chars must be digits. I mean; its a for loop and a switch.

_but_, not the end of the world.

PR #3963, opened 2025-03-28
Fix federated routing header

### LINKED ISSUE #3962: Scheduler SDK needs to use "micro_" in routing key
When the scheduler SDK uses local federation; its not including the micro_ part in the routing key it generates for itself.

```
"headers":[{"key":"X-BEAM-SERVICE-ROUTING-KEY","value":"Steve:tests-macbook-pro_26f9e7ca959dccba898ccec36fa97b2e"}]
```

Should be "micro_Steve"

Leinnan:
Fixes #3962


cdhanna (approved):
this looks good; but
1. need a changelog entry, 
2. re-run those tests; looks like a license failure.

Leinnan:
@cdhanna  It's wrong one 🙈  I will create second PR in a second

Issue #3836, opened 2024-12-19
Stats PSDK not refreshing when stats are set server-side


allister-beamable:
## Describe the Bug 

When I do `await ctx.Stats.Refresh();` before interacting with the Stats collection, I expect the stats to reflect the latest and greatest values of those stats, but if the stat were changed server-side, the Stats collection contains stale data until I stop and start the app.

### Steps to reproduce the behavior:

1. Set a stat to a certain value.
2. Iterate over the `Stats` collection to print the value.
3. Use C#MS code to change the stat from step 1 to a different value.
4. Call `Stats.Refresh()` and iterate over the collection to print values again.
5. Observe the value of the stat.

### Expected Behavior

* The value in the `Stats` collection should reflect the change from step 3 of the repro steps.

### Actual Behavior

* The value in the `Stats` collection shows the old value from step 1 of the repro steps.

#### Client Code

These calls are hooked to buttons in my test harness.

```csharp
private async Task SetRandomStat()
{
	var client = new BoxMicroserviceClient();
	var theNumber = await client.SetNumericStat();
	Debug.Log($"Numeric stat setting completed: {theNumber}");
}

private async Task FetchStats()
{
	var ctx = await BeamContext.Default.Instance;
	await ctx.Stats.OnReady;

	Clear();
	Debug.Log("Reading current stats...");
	await ctx.Stats.Refresh();
	foreach (var statPair in ctx.Stats)
	{
		Debug.Log($"S: {statPair.Key}={statPair.Value}");
	}
}
```

#### Server Code

```csharp
[ClientCallable, SwaggerCategory("StatsTests")]
public async Task<int> SetNumericStat()
{
	var objectId = $"client.public.player.{Context.UserId}";
	var theNumber = Randomizer.Next();
	var statsRequest = new NumericStatsUpdateRequest
	{
		set =
		{
			["NUMERIC_SOMETHING"] = theNumber
		}
	};
	var uri = $"/object/stats/{objectId}";
	var response = await Requester.Request(Method.POST, uri, statsRequest, parser: s => s);
	Debug.Log($"RESPONSE: {response}");
	return theNumber;
}
```

## Screenshots

![Screenshot 2024-12-19 at 17 33 43](https://github.com/user-attachments/assets/9c4d90ce-bcaf-4b28-9f84-3934af9911a5)

## Metadata

- SDK version: Beamable Unity SDK 1.19.23
- Engine version: Unity 2021.3.27f1
- Operating System: macOS

[...]

PR #3968, opened 2025-04-01
Fix - Stats PSDK not refreshing when stats are set server-side. [3836]

DiasAtBeamable:
Fixed an issue when trying to refresh the PlayerStats client SDK. It uses an internal cache to get the player stats, which were not being updated during the refresh. Now, when performing a refresh the cache is cleared to force get the updated stats.

# Ticket
- https://github.com/beamable/BeamableProduct/issues/3836


cdhanna (approved):
nice- simple :+1:

Issue #3915, opened 2025-02-28
Rename protectedPlayerStat functions to privatePlayerStat, and enum access patterns


cdhanna:
In the C#MS SDK, we expose `Services.Stats.GetProtectedPlayerStats()`. 
We should deprecate these methods and replace them with a redesigned stat SDK to aligns closer to our docs, and we should write those docs to be in line with the new SDK. 

- Add an overload to the GetStats function that accepts new enums for Domain (game / client), and Access (public / private). 
- Add the "protected" methods to use private; mark the old ones as `[Obsolete]`.

Issue #3912, opened 2025-02-25
Stats PSDK needs to get client.private stats


cdhanna:
The PSDK only gets client.public stats, which is only half of the stat groups we need! 

The `PlayerStats` `PerformRefresh()` function is the intro; we need to add the call for the private stats. 
Ideally, we should shift this to use AutoGen instead of PlatformSubscribable under the hood.

cdhanna:
related to https://github.com/beamable/BeamableProduct/issues/3915

cdhanna:
also related to the refresh bug https://github.com/beamable/BeamableProduct/issues/3836

allister-beamable:
Reminder, ping @allister-beamable to relay the update to interested Game Makers when this fix is live.

PR #3973, opened 2025-04-03
Refactor - StatsService and StatsMicroservice for better naming and support to Private Client on StatsService

DiasAtBeamable:
# Ticket

https://github.com/beamable/BeamableProduct/issues/3912
https://github.com/beamable/BeamableProduct/issues/3915

# Brief Description

Refactor on `IstatsApi` and `IMicroserviceStatsApi` to now have new methods to handle Stats with better naming and usability. Older methods were flagged as `Obsolete`

On PlayerStats Perform Refresh now it also updates the private stats. Also, remove includeAuthHeader as false on StatsService Request so it is possible to request the player's private stats.


cli/beamable.common/Runtime/Api/Stats/AbsStatsApi.cs
```diff
@@ -151,9 +151,74 @@ ArrayDict ConvertCriteriaToArrayDict(Criteria criteria)
...
+
+			string accessValue = ParseAccessValue(access);
+			return $"{domainValue}.{accessValue}.player.";
+		}
+		
+		protected string GeneratePrefix(StatsDomainType domain, StatsAccessType access, long userId)
+		{
+			return $"{GeneratePrefix(domain, access)}{userId}";
+		}
+
```
Leinnan, line 193: I would replace both of these methods with public static class with helper method, something like: 
```csharp
public static class BeamableEnumHelper {
	public static string ToStringLowerCase(this Enum e) {
		return e.ToString("F").ToLower();
	}
}
```

Leinnan:
The diff right now is too big to understand, but I really like the direction in `AbsStatsApi` and `PlayerStats`

cli/beamable.common/Runtime/Api/Stats/AbsStatsApi.cs
DiasAtBeamable: Good catch. I will change it.

DiasAtBeamable:
> The diff right now is too big to understand, but I really like the direction in `AbsStatsApi` and `PlayerStats`

Yeah, after including the AutoGenerate files, the GitHub can't handle loading the diff well. you can use this [filter ](https://github.com/beamable/BeamableProduct/pull/3973/files/d0b956e41a9d412f1e51e30cabf78447f05a4543)to ignore the auto-generated files.

cli/beamable.common/Runtime/Api/Stats/AbsStatsApi.cs
DiasAtBeamable: Done, thanks for the feedback!

Leinnan (approved):
Really cool!

PR #3981, opened 2025-04-09
Improve registration error handling

Leinnan:
Ready to merge, I've tested it mostly on Federated Identity.


client/Packages/com.beamable/Runtime/Player/PlayerAccounts.cs
```diff
@@ -1629,62 +1652,66 @@ public async Promise<RegistrationResult> AddExternalIdentity<TCloudIdentity, TSe
...
+								new ChallengeSolution
+								{
+									challenge_token = response.challenge_token, solution = solution
+								});
+						}
+						catch(PlatformRequesterException e)
+						{
+							res.innerException = e;
+							var error = PlayerErrorFromPlatformError(e.Error);
+							res.error = error == PlayerRegistrationError.OTHER_ERROR ? PlayerRegistrationError.FAILED_FEDERATED_AUTHORIZATION : error;
```
allister-beamable, line 1680: Am I the only one who has to read ternary operator expressions two or three times before understanding them? 🤣 
Just as a thought experiment, it would be more verbose to say:
```csharp
res.error = error;
if (error == OTHER_ERROR)
{
  res.error = FAILED_FEDERATED_AUTHORIZATION;
}
```
but is it easier to understand? Maybe. Maybe not. I would say keep it as it is unless you agree that an if statement is easier to read. Just wanted to make some noise about it because I always stumble on ternary operators.

allister-beamable (approved):
I like it. I am a big fan of datatypes for results that encompass both successful and failed results and "pushing the smarts to the edges" on deciding what to do with success or failure.

client/Packages/com.beamable/Runtime/Player/PlayerAccounts.cs
Leinnan: @allister-beamable I've added a parameter to helper method now, do you find it more readable now? :D

allister-beamable: I like the new way a lot! It makes it easy to read and guess your intent in the main flow of logic, and then if anyone is curious about what "PlayerErrorFromPlatformError" means in every specific case, they can read the body of that method. 10/10 would witness refactoring again!

gabrielbeamable (approved):

PR #3982, opened 2025-04-10
Helper menu option for updating Realm configuration for GPGS integration

Leinnan:
This could be really useful for game makers. Right now they need themself get the file, convert it into Base64 (most often it will end up on some third party website), update the correct key and also make sure that it is the same key that they have configured in theirs GPGS plugin configuration. 

I am not so sure if that is the right place in repo for that, so I am open to suggestions.


Leinnan:
> testing: how has this been tested, and is there a way to automate future testing of it?

I was testing it a lot while helping out Andree from Discord. Basically this PR is result of my frustration with what we have currently 😅 

> documentation: doesn't have to be in ReadMe, but at least a GitHub Wiki page describing when you'd want to use this menu item, how to do it, and where to look for the file that needs converting

I was thinking about it, and since you are mentioning this as well I will prepare one and add it to this PR

Leinnan:
@allister-beamable docs are in place

cdhanna (approved):
this seems wonderful <3

allister-beamable (approved):
With the docs in place, I approve!

PR #3988, opened 2025-04-14
fixes #3795 - C#MS Signed Requester

### LINKED ISSUE #3795: Add TrackEvent method to Microservice SDK
### Describe your Feature Request

The client-side Unity SDK has a `TrackEvent` method for emitting custom analytics to the Beamable Analytics backend. It would be nice to have an ergonomically similar method in the C#MS SDK. This Gist shows how to do a simple custom analytics event using signed requests: https://gist.github.com/allister-beamable/ece03fb78f5f8cc340eb77b54b911c22

The reason that analytics emission needs to do all this is because of how the Beamable gateways are arranged: there is a separate gateway for analytics, with special routing rules at the load-balancer level to send `api.beamable.com` requests to the proper gateway depending on path. However, C# Microservice requests to the core API follow a different path: they bypass the load balancer and go directly to the main gateway. Thus, in order to reach any gateway other than the main one, requests have to go in "by the front door" which is what the signed request stuff in the above Gist is doing.

### Describe the Desired Solution

There should be a `TrackEvent` method in the C# Microservice SDK that has exactly the same calling conventions as `TrackEvent` in the Beamable Unity SDK (or if that cannot happen, as close as possible). Something like `Services.Analytics.TrackEvent(new CoreEvent { /* ... core event stuff */ });`

It should be easy to use `TrackEvent` in a fire-and-forget fashion: usually analytics events are meant to be emitted quickly and without impeding the flow of normal code execution. Even if the method is async, the default mode should be to not await it.

### Describe Alternatives Considered

The Gist shows that it is possible to hand-write a request that emits a custom analytics event, but it is clunky and requires detailed knowledge of the precise form of event JSON. Furthermore, it requires the realm secret to be known in "user-space" code; ideally the SDK should handle authentication for me automatically, only exposing the gory details if/when I choose to "drop down a level".

### Additional context

See also https://github.com/beamable/BeamableProduct/issues/3661

cdhanna:
adds a signed requester to the C#MS. 
By default it will adopt the playerId of the callable, but it can be cleared out if required. 

Also swaps the Analytics server to use the signed requester.


cli/cli/Utils/CliSignedRequesterConfig.cs
```diff
@@ -0,0 +1,42 @@
...
+        try
+        {
+            var res = await _realmsApi.GetAdminCustomer();
+            var project = res.customer.projects.FirstOrDefault(p => p.name == _appContext.ProjectName);
+            return project.secret;
+        }
+        catch (RequesterException ex) when (ex.Status == 401)
+        {
+            throw new CliException(
+                message: $"The authorization context is not sufficient to in the current realm=[{_appContext.ProjectName}]. " +
```
gabrielbeamable, line 37: ... is not sufficient to <> in the current realm

Seems like something is missing in this part of the message

gabrielbeamable (approved):

PR #4022, opened 2025-04-25
BeamableBehaviour inspector portal button

Leinnan:
Available in play mode only:

image


cdhanna (approved):
Don't love the UIElements at this point; but, very cool regardless 💣

PR #4029, opened 2025-04-30
Minor Content Window improvements

Leinnan:
I know that there is a upcoming redesign, but before that happens I wanted to fix some pain points with selecting content in current window.

It was especially annoying when you had one item selected, then you switched content group to other where is only one content, new content is selected in the list, but in inspector you still have old content and in order to switch to new content you need to go to `Show all` and then select new content again. Now this issue is gone. As bonus, it tries to preserve selected content on filter change, but only for content that matches new filter.

https://github.com/user-attachments/assets/4371deb8-685d-4ae5-b7cc-1a30a6c12a3d


allister-beamable (approved):
Looks like it simplifies things and makes them better. Double win!

I would like to see a review from Chris or another teammate before merging, since I am less familiar with this stretch of code than y'all.

PR #4036, opened 2025-05-08
Add ToString method to OptionalString for easy print debugging.

allister-beamable:
# Ticket

No ticket, this is just a small thing that I tripped over during other debugging.

# Brief Description

Adds a type-specific ToString method to `OptionalString` so that it is easy to render optional strings in Debug.Log statements and the like.

# Checklist

# Notes


cdhanna (approved):
I'm fine with this going in- but I would raise the sticky question about `HasNonEmptyValue` vs just `HasValue`. 

An optional value of `null` is semantically different than a NONE. 

Maybe the ToString() should be a bit more explicit about when the optional value _is_ set, but has a null/empty string _value_ ?

PR #4051, opened 2025-05-19
Refactor - Switched ICallableGenerator usage to a reusable method.

DiasAtBeamable:
Refactored the ICallableGenerator to a reusable method to check for Federation methods. Now it doesn't need to check all methods for all Federations, reducing the loop iterations to a single array loop instead of one full loop for each Federation.


microservice/beamable.tooling.common/Microservice/ServiceMethodHelper.cs
```diff
@@ -43,19 +40,80 @@ public static class ServiceMethodHelper
...
       {
          var output = new List<ServiceMethod>();
          foreach (var provider in serviceMethodProviders)
          {
             output.AddRange(ScanType(serviceAttribute, provider));
-            foreach (var gen in generators)
+            if (provider.instanceType == typeof(AdminRoutes))
             {
-	            output.AddRange(gen.ScanType(serviceAttribute, provider));
+	            continue;
```
cdhanna, line 51: this is just an optimization, right? It would theoretically not break anything to scan the `AdminRoutes` class, because it does no federation anyway, correct?

DiasAtBeamable: Yeah, the idea is to skip the AdminRoutes during the for loop to reduce unnecessary checks later.

cdhanna (approved):

PR #4072, opened 2025-05-29
Improve IAP error handling

Leinnan:
When Exception cannot be casted to ErrorCode, try casting to `PlatformRequesterException` first.


allister-beamable (approved):
Ship it.

PR #4073, opened 2025-05-29
Otel - Collector error handling

gabrielbeamable:
# Brief Description

This does the following:
 - Creates a buffer with the collector logs and sends it in the UDP broadcast
 - Shows the logs in case something wrong happens with the collector, as part of the error message
 - Inside the collector, pings clickhouse using the credentials passed through environment variables to test if the connection is going to work, otherwise crashes the collector
 - In case a microservice tries to run without a docker engine running, then it will add to the error log that this was the reason for crashing


microservice/beamable.tooling.common/Microservice/CollectorManager.cs
```diff
@@ -370,7 +399,11 @@ public static async Task<int> StartCollector(
 			}
 			catch (Exception ex)
 			{
-				logger.LogError($"Collector discovery failed! ({ex.GetType().Name}) {ex.Message}\n{ex.StackTrace}");
+				logger.LogError($"Collector discovery failed! ({ex.GetType().Name}) {ex.Message}\n{ex.StackTrace} All Collector Logs=[\n{
+					string.Join("\n", CollectorStatus.logs.Select(l => $"{l.Level} - {l.Message}").ToList())
```
gabrielbeamable, line 403: Do a order by the date before printing the string here

microservice/beamable.tooling.common/Microservice/CollectorManager.cs
```diff
@@ -509,18 +540,22 @@ public static async Task<CollectorStatus> IsCollectorRunning(Socket socket, Canc
 					continue;
 
 			}
+			var logs = collector.logs.Select(l => new CollectorLogEntry() { Level = l.Level, Message = l.Message }).ToList();
```
gabrielbeamable, line 543: do the order by in here

PR #4074, opened 2025-05-30
DummyPurchaser and new TryGetLocalizedPrice method

Leinnan:
Change result:
- When no `IBeamablePurchaser` is provided it will no longer hang forever on `await BeamContext.Default.Api.BeamableIAP`. 
- we are providing new method for getting localized price, this one also right away gives information if the data was found. For old one you would need to know that it can return `???` when localized price was not found and then parse it based on that information.

### DIFF
--- a/client/Packages/com.beamable/Runtime/Core/Platform/SDK/Payments/DummyPurchaser.cs
+++ b/client/Packages/com.beamable/Runtime/Core/Platform/SDK/Payments/DummyPurchaser.cs
@@ -1,0 +1,32 @@
+using Beamable.Common;
+using Beamable.Common.Dependencies;
+
+namespace Beamable.Api.Payments
+{
+	/// <summary>
+	/// Dummy Purchaser that is used when there is no `IBeamablePurchaser` registered in BeamContext Service Provider.
+	/// </summary>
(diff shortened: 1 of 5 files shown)


client/Packages/com.beamable/Runtime/Core/Platform/SDK/Payments/DummyPurchaser.cs
```diff
@@ -0,0 +1,32 @@
...
+	public class DummyPurchaser : IBeamablePurchaser
+	{
+		public PurchasingInitializationStatus InitializationStatus { get; } = PurchasingInitializationStatus.ErrorPurchasingUnavailable;
+		public Promise<Unit> Initialize(IDependencyProvider provider = null)
+		{
+			return Promise<Unit>.Successful(new Unit());
+		}
+		public string GetLocalizedPrice(string skuSymbol)
+		{
+			return string.Empty;
```
cdhanna, line 18: curious; these implementations essentially act as no-ops... Do you think that is explicit enough? If they instead threw `NotImplementedExceptions`; do you think that would be better or worse for the user?

cdhanna (approved):

PR #4086, opened 2025-06-05
Fix - Open API Generator Fixed for Both Unreal and Unity.

DiasAtBeamable:
Fixes:

- Adjusted Dictionary Support for C# openApi gen.
- Adjusted Enum support for C# openApi gen.
- Skip Optional types on schema generator because Unreal generates all schemas found in openApi, and we don't need to add them.
- Fixes an Issue when the schema doesn't have a format or type, resulting in a nullref.
- Fixes byte type generator on Unreal Code Generator
- Fixes the Enum generation on Unreal Code Generator.


cdhanna (approved):
seems fine to me, but probably someone from the UE land should comment as well.

PR #4153, opened 2025-07-08
fixes #3986 - heartbeat removal

### LINKED ISSUE #3986: When "beamable" notifications are used (via websocket), we should no longer send heartbeat requests from clients
## Reason for the Feature
Since we will send the heartbeats from the connected websocket, sending them from the SDK is no longer necessary and eats up CPU time unnecessarily in our API and Gateway services.

I would like for us to stop sending the requests when "beamable" notifications are configured.

cdhanna:
When the realm is using a beamable websocket, this stops the heartbeats from firing all the ding dang time.


client/Packages/com.beamable/Runtime/BeamContext.cs
```diff
@@ -675,14 +681,14 @@ async Promise SetupGetUser()
...
 
-			async Promise SetupNewSession()
+			async Promise SetupNewSession(RealmConfiguration config)
 			{
 				var adId = await AdvertisingIdentifier.GetIdentifier();
 				var promise = _sessionService.StartSession(AuthorizedUser.Value, adId);
 				await promise.RecoverFromNoConnectivity(_ => new EmptyResponse());
-				if (CoreConfiguration.Instance.SendHeartbeat)
+
+				if (config.IsUsingPubnubNotifications() || CoreConfiguration.Instance.SendLegacyHeartbeat)
```
cdhanna, line 690: this is too restrictive.

PR #4157, opened 2025-07-09
Feature/beam 4120

molaycule:
# Ticket

https://disruptorbeam.atlassian.net/browse/BEAM-4120

# Brief Description

> Signed Request Implementation
> Leaderboard Service
> New `BeamServer` Class

# Checklist


web/src/utils/makeApiRequest.ts
```diff
@@ -5,13 +5,13 @@ import { HttpRequest } from '@/network/http/types/HttpRequest';
...
-  m: HttpMethod;
-  q?: Record<string, unknown>;
-  p?: TReq;
-  g?: string;
-  w?: boolean;
+  r: HttpRequester; // The HTTP requester instance to use for making the API request
+  e: string; // The endpoint URL to which the request will be sent
+  m: HttpMethod; // The HTTP method to use for the request (e.g., GET, POST, PUT, DELETE)
+  q?: Record<string, unknown>; // Optional query parameters to include in the request URL
+  p?: TReq; // Optional payload to include in the request body (for POST, PUT, etc.)
```
cdhanna, line 12: curious; why the tiny variable names?

cdhanna (approved):

web/src/utils/makeApiRequest.ts
molaycule: For weird reasons, the bundler was always spitting out the full name e.g e = endpoint, m = method, q = queries, p = payload, because I think it see it as object props, so the final bundle always include the text endpoint etc in full for all the api methods even in minified bundles. But I am trying to see if switching to pure functions for the apis and not classes with help improve this

PR #4326, opened 2025-10-02
Fix: Content Window issues

DiasAtBeamable:
- Issue when trying to rename a Content on a virtual scrollview
 - Issue when searching for items that the content is content subtype
 - Issue when contents are updated, but the Content Editor is not repainted.


cdhanna (approved):
👍

PR #4443, opened 2025-12-09
Improve open api schema population

### LINKED ISSUE #4407: OAPI schema generation can accidentally add the same type twice to a dictionary
And dictionaries cannot have the same key twice.

Leinnan:
Resolves #4407

Besides that it also improves OpenAPI components generation flow, so for example when we have class like `MicroserviceRuntimeMetadata` with array field `public List<FederationComponentMetadata> federatedComponents = new List<FederationComponentMetadata>();` it detects that it is a separate class that is and stores information about the fact that the schema should include type information for `FederationComponentMetadata` as well. In old version the schema for the field was storing only that:
```
          "federatedComponents": {
            "type": "array"
          }
```
Which only informs that there is array of anything. New version gives information about items type and also stores schema for that field type.


allister-beamable (approved):
The principle seems sound! I am not too familiar with the nuts and bolts of this code, so I defer to Chris and Dias for that part of the review :)

DiasAtBeamable (commented):
The code looks good to me. But I cannot confirm whether some of the changes will affect how the ClientCode is generated, primarily on the Unreal side. @PedroRauizBeamable has some Unreal project that would be good to test it. I have a sample C# Microservice Code with a lot of callables to test a bunch of types of usages. Let me know if you want it, I can send it to you.

allister-beamable:
@Leinnan can you check this over after rebasing or merging `main` and see whether any of Copilot's comments need addressing?

PR #4446, opened 2025-12-11
Microservice build- Specify the output path for the app binaries

Leinnan:
# Issue

When I did create a new beam project from the CLI it all is working fine, I can create new service and `dotnet beam deploy plan` works fine. But when after creating new storage for that service docker builds stops working. I did a digging and find out that in my example the command that was run for build was:
```
dotnet publish BeamableServices.sln --verbosity minimal --no-self-contained --runtime unix-x64 -p:BeamPlatform=lin -p:BeamRunningArchitecture=x64 -p:BeamCollectorPlatformArchArg="--platform lin --arch x64"  --configuration Release -p:Deterministic="True" -p:BeamGenProps="disable" -p:GenerateClientCode="false" -p:CopyToLinkedProjects="false" -logger:"cli.Utils.MsBuildSolutionLogger,/Users/test/.nuget/packages/beamable.tools/6.2.1/tools/net9.0/any/Beamable.Tools.dll" -p:PublishDir="bin/beamApp/support" -p:BeamPublish="true"
```

Result of that command was that `services/ForgeService/bin/beamApp/support` got populated, but `services/ForgeService/bin/beamApp/app` was empty.  Adding `-p:OutputPath="bin/beamApp/app"` to that command fixed the problem. 

I assume this is the issue some of our customers are facing from time to time when they get `chmod: /beamApp/{ServiceName}.dll: No such file or directory` errors. At least this was my case.

### DIFF
--- a/cli/cli/Commands/Project/BuildSolutionCommand.cs
+++ b/cli/cli/Commands/Project/BuildSolutionCommand.cs
@@ -28,10 +28,10 @@ public class BuildSolutionCommandResults
 }
 public class BuildSolutionCommand : StreamCommand<BuildSolutionCommandArgs, BuildSolutionCommandResults>
 {
-    
+
     // I haven't thought too hard about this yet; this command
     //  was developed as a way to test solution-level building
-    //  for usage in the plan/release flows. 
+    //  for usage in the plan/release flows.
     public override bool IsForInternalUse => true;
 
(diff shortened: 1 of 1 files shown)


Leinnan:
The issue was simpler- it was using wrong dotnet. Information for future me- if that happens make sure that project contains `global.json` in the root directory with dotnet version specified.

PR #4452, opened 2026-01-08
fixes #4245

### LINKED ISSUE #4245: Validate- what happens in the CLI log in flow if your first dev realm is archived?
This looks fine, but customer reports issue where Unity always tries to log into an archived realm.

```csharp

	private async Task<string> PickGameAndRealm(InitCommandArgs args)
	{
		var didLogin = await Login(args);
		if (!didLogin)
		{
			return string.Empty; // cannot fetch games without correct credentials
		}
		var games = await _realmsApi.GetGames().ShowLoading("Fetching games...");
		var gameChoices = games.Select(g => g.DisplayName.Replace("[PROD]", "")).ToList();
		var gameSelection = args.Quiet ? gameChoices.First() : AnsiConsole.Prompt(
			new SelectionPrompt<string>()
				.Title("What [green]game[/] are you using?")
				.AddChoices(gameChoices)
				.AddBeamHightlight()
		);
		var game = games.FirstOrDefault(g => g.DisplayName.Replace("[PROD]", "") == gameSelection);

		var realms = await _realmsApi.GetRealms(game).ShowLoading("Fetching realms...");
		var realmChoices = realms
			.Where(r => !r.Archived)
			.Select(r => $"{r.DisplayName.Replace("[", "").Replace("]", "")} - {r.Pid}");
		var realmSelection = args.Quiet ? 
			realms.Where(r => r.Depth == 2).OrderBy(r =>r.Pid).Select(r => $"{r.DisplayName.Replace("[", "").Replace("]", "")} - {r.Pid}").First() :
			AnsiConsole.Prompt(new SelectionPrompt<string>()
				.Title("What [green]realm[/] are you using?")
				.AddChoices(realmChoices)
				.AddBeamHightlight()
		);
		var realm = realms.FirstOrDefault(g => realmSelection.StartsWith(g.DisplayName.Replace("[", "").Replace("]", "")));
		return realm.Pid;
	}
```

cdhanna:


cli/cli/Commands/InitCommand.cs
```diff
@@ -384,30 +384,53 @@ private async Task<string> PickGameAndRealm(InitCommandArgs args)
...
 		var games = await _realmsApi.GetGames().ShowLoading("Fetching games...");
 		var gameChoices = games.Select(g => g.DisplayName.Replace("[PROD]", "")).ToList();
-		var gameSelection = args.Quiet ? gameChoices.First() : AnsiConsole.Prompt(
-			new SelectionPrompt<string>()
-				.Title("What [green]game[/] are you using?")
-				.AddChoices(gameChoices)
-				.AddBeamHightlight()
-		);
+		var gameSelection = args.Quiet
+			? gameChoices.First()
```
allister-beamable: Just thinking out loud: there is no way for the first game to be 100% archived, right? That is, because we disallow archiving top-level game prod realms, there will always be at least 1 realm in the first game? If so, all is well.

PR #4463, opened 2026-01-14
fixes #4461 - lower log level

### LINKED ISSUE #4461: Change docs logging to Trace level, so they can be excluded while retaining valuable Debug level logs
## Reason for the Feature
When accessing the C# MS page in portal, viewing docs, etc, the admin docs API is called repeatedly for a C# MS.  This causes a TON of Debug level logging to happen, writing out a line per route, etc and it's spammy making it impossible to find other, valuable, debug logs. 

Change the log level of the docs logs to Trace, so we can exclude them and retain the valuable debug logs.

## Alternatives or Workarounds
No workarounds, we need other debug logs, just not these docs logs.

cdhanna:


allister-beamable (approved):
Yes, please! Cleaner logging is nicer logging ☀️

PR #4562, opened 2026-04-06
Feature - Added support to run detached portal extensions.

DiasAtBeamable:
This also refactors how we run Processes in CLI, reducing boilerplate code for OS specific or detached-specific options.

### DIFF
--- a/cli/cli/Commands/Project/RunProjectCommand.cs
+++ b/cli/cli/Commands/Project/RunProjectCommand.cs
@@ -132,13 +132,13 @@ public partial class RunProjectCommand : AppCommand<RunProjectCommandArgs>
 		// First, we need to find out which services are currently running.
 		if (args.forceRestart)
 		{
-			Log.Verbose("starting discovery");
+			Log.Trace("starting discovery");
 			await StopProjectCommand.DiscoverAndStopServices(args, new HashSet<string>(args.services), "project run command", kill: true, TimeSpan.FromMilliseconds(100),
 				evt =>
 				{
 					// do nothing?
 				});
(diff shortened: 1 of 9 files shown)


cli/cli/Commands/Project/RunProjectCommand.cs
```diff
@@ -463,6 +512,109 @@ public static async Task RunService(
...
+			fullInnerCommand += $" --require-process-id {args.requireProcessId}";
+		}
+
+		string exe;
+		string commandStr;
+
+		// Get the correct command for each OS
+		if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+		{
+			commandStr = "/C " + fullInnerCommand.EnquotePath('(', ')');
```
gabrielbeamable: I know it's not the scope of this task, but we really should centralize how we start/kill processes. I think this same logic is spread across this project and that makes me sad 😿

DiasAtBeamable: Yeah, it is almost the same logic as we use when running detached microservices. 
The difference is that the microservice has some changes when running detached or not. 

Let me think of a way to create a Util function for them.

By any chance, do you remember any other place that also calls a process in the same way?

DiasAtBeamable: Done, I did a refactor to reuse a bunch of OS specific and detach-specific code inside the StartProcessUtil.Run
Asked AI to create a summary for it, seems good enough.

gabrielbeamable (approved):

PR #4570, opened 2026-04-14
Feature - Track portal extension performance

DiasAtBeamable:
Tracks through Otel traces the following:
 - Build Time
 - Install Time
 - Build Size for each file
 - Build Total Size


cli/cli/Services/BeamoLocalSystem_PortalExtension.cs
```diff
@@ -61,10 +69,14 @@ await BeamServer
...
 						var attributes = provider.GetService<MicroserviceAttribute>();
-
-						observer.ConfigureServiceData(notification, attributes, localSystem.BeamoManifest);
+						observer.ConfigureServiceData(notification, attributes, beamActivity, localSystem.BeamoManifest);
 						observer.InstallDeps();
 						observer.BuildExtension();
+						
+						// We can dispose beamActivity here now as we are only getting the Install and Build traces, as the 
+						// Portal Extension run forever until we force it to stop, we don't need wait for that
+						beamActivity.Dispose();
```
gabrielbeamable, line 75: this got me a bit confused, this is the main PE activity, right? So if you dispose of it here, it won't exist later for the rest of the execution of this command. So, when it saves this to file, does the main activity appears there? and does the child ones with the spans for build time have the main activity as parent id?

gabrielbeamable:
It would be nice to get some of the data published to our dev clickhouse, so we can see it showing in Grafana for example

cli/cli/Services/BeamoLocalSystem_PortalExtension.cs
DiasAtBeamable: So, without this, we never get the Activity trace for the run command. As the CLI command starts an MS with the RunForever command, it will never dispose of the activity because the only way to stop this is by doing Ctrl/Cmd + C or by a `dotnet beam project stop`. When using Ctrl/Cmd + C, the process stops immediately and does not dispose of the activity. So the best place that I found to get the main activity for what matters, which is starting the portal extension service, is after the install and build.

PR #4572, opened 2026-04-15
Fix - Null values for ContentObject fields after creation

### LINKED ISSUE #4569: Content ValidationAttribute issues
## Describe the Bug 

Creating a custom ValidationAttribute that checks for the issues in the content can result in multiple issues.
One was that when I created ValidationAttribute that was not passing for default content I no longer can create that content from the Unity editor. A

## To Reproduce

Code:
```csharp
using System.Collections;
using System.Collections.Generic;
using Beamable.Common.Content.Validation;
using Beamable.Common.Inventory;
using UnityEngine;

namespace Test
{
    
    public class ValidateStuff : ValidationAttribute
    {
        private const string FEDERATION = "MyService/providerId";
        public override void Validate(ContentValidationArgs args)
        {
            // check if collection name is allowed
            var content = args.Content;
            if (content is not ItemContent nftItemContent)
                return;
            
            var validationField = args.ValidationField;
            Debug.Log("Is null:" + (nftItemContent.federation == null));

            // first step is to ensure this is federated
            if (!nftItemContent.federation.HasValue ||
                nftItemContent.federation.Value.Namespace != FEDERATION)
            {
                throw new ContentValidationException(nftItemContent, validationField,
                    $"All nft items must be federated by MyService. Select federation and {FEDERATION}");
            }

            var value = validationField.GetValue().ToString();

            if (string.IsNullOrEmpty(value)) {
                throw new ContentValidationException(nftItemContent, validationField,
                    "Collection name cannot be empty");
            }
        }
    }
}
```

```csharp
using Beamable;
using Beamable.Common.Content;
using Beamable.Common.Inventory;
using UnityEngine;

[System.Serializable]
public class CustomItemRef : ItemRef<CustomItem> { }

[ContentType("CustomItem")]
[System.Serializable]
public class CustomItem : ItemContent
{
    [Test.ValidateStuff]
    public string SuperName;
    // public string SuperName = "DSA";

}
```

DiasAtBeamable:
Fixed an issue when creating a new Content Object some Optional and string values were null instead of default values.

### DIFF
--- a/client/Packages/com.beamable/CHANGELOG.md
+++ b/client/Packages/com.beamable/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Deserialization issue with `properties` field in Score Items of Events
 - Fixed an issue where the Unity Editor would not detect changes to Icon subObject (for Sprites in Multiple Mode) and thus not saving it correctly
+- Fixed an issue when creating a new Content Object some Optional and string values were null instead of default values.
 
 ### Changed
 - Added support to configure Max Parallel Service Build count on MicroserviceConfiguration.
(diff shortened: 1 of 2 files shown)


Leinnan (approved):
👍

cdhanna (approved):

PR #4580, opened 2026-04-23
Change - Enhance content publish command performance with size-aware batching and retry logic

### LINKED ISSUE #4579: Consider performance impact of 20-item batch size in content publish
## Summary

`dotnet beam content publish` (CLI v7.x) chunks content into batches of 20 items (`CONTENT_PUBLISH_BATCH_SIZE` in `CliConstants`) and fires all batches concurrently via `Task.WhenAll`. For a catalog of 4500+ items at roughly 2 KB apiece, this produces over 200 simultaneous POST requests to `/basic/content`.

## Background

The batch size was introduced to avoid hitting the gateway's 8 MiB request body limit (which rejects the single-POST approach used in v4.x). At 20 items × ~2 KB, each individual request is well within limits (~40 KB). However, firing 200+ requests in parallel may stress connection pools and the gateway in ways that are harder to predict than a single oversized body.

## Concern

- A catalog of N items produces ⌈N/20⌉ concurrent requests, all launched at once
- For a 4500-item catalog that is ~225 concurrent POSTs
- There is no throttling, back-pressure, or concurrency cap on the batch dispatches
- `CONTENT_PUBLISH_BATCH_SIZE` is a compiled-in `const`; customers cannot tune it without rebuilding the CLI

## Suggestion

Consider either:
1. Increasing the batch size (e.g. 500-2000 items keeps well under the 8 MiB limit while reducing concurrency to 3-10 requests), or
2. Processing batches sequentially, or with a bounded concurrency (e.g. `SemaphoreSlim`)

A 2000-item batch at ~2 KB per item is ~4 MB — 50% of the gateway limit and a comfortable margin.

DiasAtBeamable:
Issue: https://github.com/beamable/BeamableProduct/issues/4579

### DIFF
--- a/cli/cli/Services/Content/ContentService.cs
+++ b/cli/cli/Services/Content/ContentService.cs
@@ -74,6 +74,41 @@ public partial class ContentService
 	private const int ERR_CODE_PUBLISH_FAILED_INVALID_REFERENCE_MANIFEST = 3;
 
 	/// <summary>
+	/// Target POST body size for each content-publish batch.
+	/// Set to 1 MiB — 12.5 % of the gateway's 8 MiB hard limit — to keep
+	/// individual requests well within the rejection threshold for catalogs of
+	/// large content items, while still consolidating many small items into
+	/// fewer requests than the old fixed-count approach.
+	/// </summary>
+	private const long CONTENT_PUBLISH_TARGET_BATCH_BYTES = 1 * 1024 * 1024; // 1 MiB
+
+	/// <summary>
+	/// Maximum number of content items per batch POST, regardless of their
+	/// serialized byte size.  Caps the amount of server-side work in a single
+	/// request so that the content service does not time out while processing
+	/// a batch of many tiny items that would otherwise be well under the byte
+	/// limit.  Both <see cref="CONTENT_PUBLISH_TARGET_BATCH_BYTES"/> and this
+	/// constant must be satisfied simultaneously: a batch is flushed as soon
+	/// as either threshold is reached.
+	/// </summary>
+	private const int CONTENT_PUBLISH_MAX_ITEMS_PER_BATCH = 50;
+
+	/// <summary>
+	/// Maximum number of content-publish POST requests that may be in flight
+	/// at the same time.  Limits connection-pool pressure and gateway load
+	/// regardless of how many batches the size-aware splitter produces.
+	/// </summary>
+	private const int CONTENT_PUBLISH_MAX_CONCURRENCY = 8;
+
+	/// <summary>
+	/// How many times a single batch POST may be retried after a transient
+	/// gateway error (429, 502, 503, 504) before the whole publish is aborted.
+	/// Delays follow an exponential back-off of 2^(attempt+1) seconds
+	/// (2 s → 4 s → 8 s) with ±20 % random jitter.
+	/// </summary>
+	private const int CONTENT_PUBLISH_MAX_RETRIES = 3;
+
+	/// <summary>
 	/// <see cref="CreateFakeEmptyManifest"/> and <see cref="GetManifest"/>.
 	/// </summary>
 	private const string FAKE_EMPTY_MANIFEST_UID = "EmptyManifest";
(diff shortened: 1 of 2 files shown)


cli/cli/Services/Content/ContentService.cs
```diff
@@ -752,6 +787,103 @@ public async Task BulkSaveLocalContent(LocalContentFiles localCache, string[] co
...
+	/// new one is started.  An item that is larger than
+	/// <paramref name="targetBytes"/> on its own is always emitted as a
+	/// single-item batch rather than being dropped.
+	/// </summary>
+	private static IEnumerable<ContentDefinition[]> BuildSizeAwareBatches(
+		IEnumerable<ContentDefinition> items, long targetBytes, int maxItemsPerBatch)
+	{
+		// Overhead of the SaveContentRequest wrapper: {"content":[…]}
+		// 13 bytes regardless of the number of items.
+		const int wrapperOverheadBytes = 13;
```
allister-beamable: Is the wrapper overhead even needed in the calculation, as long as we aren't flying too close to the ceiling? Like, if the target size is 1 MiB, that is _waaay_ less than the 8 MiB limit in gateway, so 13 bytes is absolutely not going to matter.

cli/cli/Services/Content/ContentService.cs
```diff
@@ -752,6 +787,103 @@ public async Task BulkSaveLocalContent(LocalContentFiles localCache, string[] co
...
+
+		var batch = new List<ContentDefinition>();
+		long batchBytes = 0;
+
+		foreach (var item in items)
+		{
+			// Serialise the item in isolation to get a byte-accurate estimate.
+			var itemJson = JsonSerializable.ToJson(item);
+			var itemBytes = Encoding.UTF8.GetByteCount(itemJson);
```
allister-beamable, line 854: What is the time cost of this computation? For very large bodies of content, I worry that this might end up costing double-digit seconds of computation (depending on developer machine strength). Worth measuring, at the very least.

allister-beamable: Though it occurs to me, if we have to JSONify things for inclusion in the batch anyhow, doing it early to get an estimate is no additional cost. 🤔

allister-beamable (changes requested):
I like the idea of this, but have mild concerns about any time overhead it might add. I would love to see time measurements (before and after, if possible!) with realistic bodies of content.

cli/cli/Services/Content/ContentService.cs
DiasAtBeamable: Yeah, that's true, good call. I'll remove it.

cli/cli/Services/Content/ContentService.cs
DiasAtBeamable: I ran a benchmark test for this. We're fine adding this extra serialization. 

Here are the results:

| Size | Items | Total Payload | Serialization Time | Per Item | Throughput |
|------|------:|-------:|-------:|-------:|-------:|
| **Small** (~218 B) | 50 | 0.01 MiB | 0.1 ms | 0.001 ms | 146.6 MiB/s |
| **Small** | 500 | 0.11 MiB | 0.8 ms | 0.002 ms | 128.0 MiB/s |
| **Small** | 5,000 | 1.08 MiB | 7.1 ms | 0.001 ms | 151.2 MiB/s |
| **Medium** (~10 KB) | 50 | 0.50 MiB | 0.2 ms | 0.003 ms | 3,172.6 MiB/s |
| **Medium** | 500 | 4.98 MiB | 2.1 ms | 0.004 ms | 2,359.6 MiB/s |
| **Medium** | 5,000 | 49.77 MiB | 32.5 ms | 0.007 ms | 1,529.5 MiB/s |
|  **Large** (~100 KB) | 50 | 4.89 MiB | 2.6 ms | 0.052 ms | 1,882.1 MiB/s |
| **Large** | 500 | 48.92 MiB | 31.5 ms | 0.063 ms | 1,553.1 MiB/s |
| **Large** | 5,000 | 489.22 MiB | 513.3 ms | 0.103 ms | 953.1 MiB/s |
| **XLarge** (~300 KB) | 50 | 14.66 MiB | 9.7 ms | 0.193 ms | 1,515.4 MiB/s |
| **Large** | 5,000 | 489.22 MiB | 513.3 ms | 0.103 ms | 953.1 MiB/s |
| **XLarge** (~300 KB) | 50 | 14.66 MiB | 9.7 ms | 0.193 ms | 1,515.4 MiB/s |


My PC has an i9-14900HX.

What do you think?

allister-beamable: Hooray for measuring! It might be worth trying it on a mobile device: phone processors are considerably less powerful than a recent-generation i9. Still, even if the mobile processor is 50x slower, that is still less than half a second.

allister-beamable (approved):
Changes look good.

cdhanna (changes requested):
looks good- but please add changelog line

PR #4585, opened 2026-04-24
Feature/cli mcp server

### LINKED ISSUE #4430: Make CLI an MCP Server

DiasAtBeamable:
# MCP Server for Beamable CLI

## Context

The Beamable CLI (`beam`) has 200+ commands covering content, services, auth, realms, and more. Adding `beam mcp serve` starts a stdio MCP server that AI clients (Claude, VS Code, Cursor, etc.) can connect to, allowing them to invoke any `beam` command as a tool to automate Beamable workflows.

**Key design constraint:** Exposing all commands as individual MCP tools would cost tens of thousands of tokens in context per connection. Instead the server exposes exactly **5 meta-tools** (~380 tokens total). The model calls `--help` on any command path to discover subcommands and parameter schemas, reusing the CLI's built-in help system with zero custom reflection. Token overhead never grows as new commands are added.

---

## The 5 Tools

| Tool | Description |
|---|---|
| `beam_list_commands` | Call `beam [prefix] --help` to discover subcommands. Empty prefix lists all root commands. |
| `beam_get_help` | Call `beam [command] --help` to get full options and argument docs for a specific command. |
| `beam_exec` | Run any beam command — pass everything after `beam`, e.g. `"project new microservice --name Foo"` |
| `beam_get_skill` | Load a step-by-step skill guide for complex workflows. Empty string lists all skills; pass a skill name to load the full guide. |
| `beam_get_source` | Read Beamable SDK source code. Auto-detects platform/version, resolves file paths, reads and returns file content with pagination. |

### Intended workflow

```
beam_list_commands("")          → discover root command groups
beam_list_commands("project")   → find the right subcommand
beam_get_help("project new portal-extension")  → read all args before calling
beam_exec("project new portal-extension ...")  → run the command
beam_get_source("ContentObject.cs")            → read SDK source directly
```

---

### Command entry point

`beam mcp serve` is a CLI command (`McpServeCommand`) registered under `beam mcp` (`McpGroupCommand`). Both implement `IStandaloneCommand` so they work without a `.beamable` config directory.

The `.mcp.json` client config points to this command as the server process.

### Execution: in-process

Each tool call creates a fresh `App` instance, swaps `IDataReporterService` for a `CapturingReporterService` that writes to a `StringWriter`, and calls `app.RunWithSingleString()`. Both `Console.Out` and `Console.Error` are redirected to separate `StringWriter` instances per call:

- `Console.Out` → captures structured JSON output from `IDataReporterService` and `--help` text
- `Console.Error` → captures `CliException` messages and other framework errors

After each call, any stderr content is appended as a `{"error": "..."}` JSON line so errors are always visible to the AI client. Without this, `CliException` (which writes to stderr) would be silently dropped.

`Console.Out` is redirected to `TextWriter.Null` before the MCP host starts to prevent stray writes from corrupting the stdio JSON-RPC framing.

A `SemaphoreSlim(1)` in `McpToolExecutor` serializes concurrent tool calls for thread safety.

### Automatic flags injected by `EnsureLogStreams`

| Flag | Purpose |
|---|---|
| `--emit-log-streams` | Routes `Log.Information`/`Log.Error` calls through `IDataReporterService` so they appear in the MCP response alongside structured output. Without this, log messages are sent to the ZLogger file sink and the AI client never sees them. |

### IEmptyResult handling

28 CLI commands implement `IEmptyResult`, which means the framework skips `IDataReporterService.Report()` on success. Without special handling, these commands would return an empty string to the AI client — zero feedback on whether the command succeeded. `McpToolExecutor` detects empty output and emits a `{"status":"ok","command":"..."}` envelope so the AI always knows the command completed.

### Workspace resolution via MCP roots

In sandboxed environments like Claude Cowork, the MCP server needs to know the project directory but `cwd` in MCP config is not supported (silently ignored by all clients). Instead, the server uses the MCP protocol's **roots** mechanism.

`McpServerBuilder.EnsureWorkspaceFromRootsAsync(McpServer server)` is called lazily on the first tool invocation:
1. Calls `server.RequestRootsAsync()` to get workspace URIs from the client
2. If the client returns a `file://` URI pointing to an existing directory, sets `Directory.SetCurrentDirectory()` to that path
3. If roots fail (client doesn't support them), no workspace override is applied
4. Thread-safe one-time execution via `Interlocked.CompareExchange` on a static `_workspaceResolved` flag

All 5 tool methods in `BeamMcpTools` inject `McpServer server` as a parameter (auto-resolved by the MCP SDK, no `[Description]` needed) and call `EnsureWorkspaceFromRootsAsync` before executing.

### Output filtering and truncation

To reduce noise and prevent oversized responses from consuming excessive agent context:

- **Log channel suppression**: `CapturingReporterService` has a `SuppressedChannels` HashSet that drops "logs" channel messages. Structured `Log.Information`/`Log.Error` calls routed via `--emit-log-streams` are filtered out, keeping only "stream" (results) and "error" channels.
- **Output truncation**: `TruncateIfNeeded()` caps output at 32KB. If exceeded, it keeps the first 4KB (head) and last 24KB (tail), replacing the middle with a `[N characters truncated]` message. This preserves the command preamble and the most recent output (usually the important part).
- Constants: `MaxOutputLength = 32_768`, `TruncateHeadKeep = 4_096`, `TruncateTailKeep = 24_576`

### Subprocess stdin isolation

The MCP server communicates over stdio — stdin carries JSON-RPC messages from the AI client. Child processes spawned by CLI commands (via `StartProcessUtil.Run`) previously inherited this stdin pipe. Tools like `npx` and `npm` behaved differently when their stdin was a JSON-RPC pipe instead of a terminal or `/dev/null`, causing 300-second hangs due to npm's `fetchRetryTimeout`.

**Fix:** `StartProcessUtil` sets `RedirectStandardInput = true` on all child processes and immediately closes the redirected stdin. This gives every child process an empty stdin (equivalent to `/dev/null`) instead of the MCP transport pipe, making them behave identically to terminal invocations.

---

### `McpGroupCommand.cs`
Top-level `beam mcp` command group.

### `McpServeCommand.cs`
The `beam mcp serve` command. Creates a `McpServerBuilder` and runs the MCP stdio host. Skills are always served from embedded resources.

### `InstallAISkillsCommand.cs`
The `beam install-ai-skill` root command. Extracts embedded skill files to AI agent directories (`.claude/skills/`, `.cursor/rules/`, `.windsurf/rules/`, `.opencode/skills/`) with auto-detection and per-agent flags.

### `McpToolExecutor.cs`
Core execution engine:
- `ExecuteAsync(commandLine)` — redirects `Console.Out` + `Console.Error`, runs in-process, appends any captured stderr as a JSON error line, emits `{"status":"ok"}` for empty results. Output is filtered through `TruncateIfNeeded()` before returning.
- `ExecuteHelpAsync(commandPath)` — appends `--help` and delegates to `ExecuteAsync`.
- `GetSourceCode(platform, version, filePath, offset, limit)` — auto-detects SDK platform (Unity, CLI, Unreal, Web) and version from the current project directory. Resolves file paths across NuGet cache directories (content + build folders), reads file content with 64KB pagination, and validates paths are under allowed SDK directories. Returns a `JObject` with conditional fields (`content`, `totalLength`, `offset`, `hasMore`, `nextOffset`, `remaining`).
- `ResolveFilePath(filePath, sourcePath, commonPaths)` — multi-strategy path resolution: absolute → relative to sourcePath → relative to each commonPaths entry → recursive filename search across all commonPaths directories.
- `IsPathUnderAllowedDirectory(resolvedPath, sourcePath, commonPaths)` — security check ensuring resolved file paths are under known SDK directories (case-insensitive on Windows).
- `TruncateIfNeeded(output)` — caps output at 32KB with head (4KB) + tail (24KB) strategy.
- `EnsureLogStreams(commandLine)` — injects `--emit-log-streams` into every command.
- `GetSkillAsync(skillName)` — reads embedded skill markdown files from `cli/Docs/Skills/`. Returns a JSON catalog when called with an empty string, or the full markdown content of a specific skill.
- `CapturingReporterService.Report<T>()` — suppresses "logs" channel messages via `SuppressedChannels` HashSet.

### `McpServerBuilder.cs`
Builds the MCP stdio host. Contains `EnsureWorkspaceFromRootsAsync()` for lazy workspace resolution via MCP roots. Contains `BeamMcpTools` with the 5 tool methods (`beam_list_commands`, `beam_get_help`, `beam_exec`, `beam_get_skill`, `beam_get_source`) decorated with `[McpServerTool]` / `[Description]`. All tools inject `McpServer server` and call `EnsureWorkspaceFromRootsAsync` before executing.

### `McpSetupCommand.cs`
The `beam mcp setup` command. Writes a `.mcp.json` config file to the project directory. Uses `dotnet beam` (local tool) as the command so the project-local CLI version is used. Realm context is resolved at serve time from the `.beamable` workspace, not baked into the config.

### `GenerateSkillDocsCommand.cs`
The `beam generate-skill-docs` command. Renders Scriban templates from `cli/Docs/SkillTemplates/*.md.scriban` into `cli/Docs/Skills/*.md`. Builds the full command tree via `CliGenerator.GetCliContext()` and injects command metadata (options, arguments, aliases) into templates. Also injects type schema data (federation types, Unreal type mappings).

### `SkillTemplateData.cs`
Data model for Scriban template rendering. Contains `Commands` dictionary (keyed by execution path, e.g. `"project new service"`), `FederationTypes`, `UnrealTypeMappings`, and per-command `SkillCommandDetail` with `Arguments` and `Options` arrays.

---

### Purpose

In sandboxed environments like Claude Cowork, the AI agent can only read files within the workspace directory. System directories like the NuGet cache (`~/.nuget/packages/`) or Unity's PackageCache are blocked by the sandbox. The MCP server process runs outside the sandbox with full filesystem access, so `beam_get_source` reads SDK files and returns their content through tool responses — bypassing the sandbox restriction.

### Platform detection

`GetSourceCode` auto-detects the SDK platform and version from the current project directory:

| Platform | Detection method | Source location |
|---|---|---|
| Unity | `Packages/manifest.json` contains `com.beamable` | `Library/PackageCache/com.beamable@{version}/` |
| CLI/Microservice | `.config/dotnet-tools.json` contains `beamable.tools` | NuGet cache (see below) |
| Unreal | `Plugins/BeamableCore/` directory exists | `Plugins/BeamableCore/` |
| Web | `package.json` contains `@beamable/sdk` | `node_modules/@beamable/sdk/` |

### NuGet commonPaths (CLI platform)

For the CLI/Microservice platform, `GetSourceCode` builds 5 commonPaths entries covering both source code and build system files:

| Index | Package | Path | Contents |
|---|---|---|---|
| 0 | `beamable.common` | `content/sourceCode/Runtime/` | ContentObject, IContentApi, Optional, serialization |
| 1 | `beamable.tooling.common` | `content/sourceCode/` | ClientCallableAttribute, federation interfaces, storage |
| 2 | `beamable.microservice.runtime` | `content/sourceCode/` | Microservice base class, API implementations |
| 3 | `beamable.microservice.runtime` | `build/` | MSBuild .targets/.props for OAPI gen, build validation, collector resolution |
| 4 | `beamable.common` | `build/netstandard2.0/` | MSBuild .props for the common package |

### File resolution

`ResolveFilePath` uses a multi-strategy approach (in order):
1. **Absolute path** — if rooted, canonicalize and return directly
2. **Relative to sourcePath** — try `Path.Combine(sourcePath, filePath)` if sourcePath is local
3. **Relative to each commonPaths entry** — iterate and try each
4. **Filename-only search** — if no directory separators, search all commonPaths recursively via `Directory.GetFiles(entry, fileName, SearchOption.AllDirectories)`

### Security

`IsPathUnderAllowedDirectory` validates that all resolved file paths are under known SDK directories (sourcePath or commonPaths). Paths outside these directories are rejected with a warning. Comparison is case-insensitive on Windows.

### Pagination

Files are read in chunks to avoid oversized responses:
- Default chunk size: 64KB (`DefaultFileReadLimit = 65_536`)
- `limit` parameter is capped at `DefaultFileReadLimit`
- Response includes: `content`, `totalLength`, `offset`, `hasMore`, `nextOffset`, `remaining`
- Agents call back with `offset: nextOffset` to read the next chunk

### Response shape

Built using `JObject` for conditional fields:
- **Always present**: `platform`, `detectedVersion`, `sourcePath`, `commonPaths`, `hint`
- **When file requested**: `filePath` (resolved absolute path)
- **When content read**: `content`, `totalLength`, `offset`, `hasMore`, `nextOffset` (if hasMore), `remaining` (if hasMore)

---

### Purpose

AI agents make mistakes on multi-step workflows because `--help` output lacks domain knowledge — e.g., portal extension creation requires calling `list-extension-options` first, `services run/stop/deploy` are deprecated, content sync can be destructive. Skills are curated step-by-step guides the AI agent loads on-demand before attempting complex workflows.

### How it works

`beam_get_skill` reads embedded markdown files from `cli/Docs/Skills/`. The MCP tool always serves from embedded resources — it does not read from disk.

- `beam_get_skill("")` — returns a JSON array of `[{name, description}]` for all available skills
- `beam_get_skill("beam-create-portal-extension")` — returns the full markdown content of that skill
- Unknown skill name — returns a JSON error with the list of available skill names

Skills are read directly from embedded resources (no CLI command execution, no semaphore lock needed).

### Scriban template system

Skill files in `cli/Docs/Skills/` are **generated** from Scriban templates in `cli/Docs/SkillTemplates/`. Each template (`.md.scriban`) is rendered at build time by `GenerateSkillDocsCommand`, which injects:

- **Command metadata**: Option tables for any CLI command, accessed via `{{ commands["command path"] }}` (e.g. `{{ commands["project new service"] }}`). Options and arguments are extracted from the live command tree with aliases resolved (e.g. `deployment` → `deploy`).
- **Type schema data**: Federation types (`{{ federation_types }}`), Unreal type mappings (`{{ unreal_type_mappings }}`).

Templates use PascalCase-to-snake_case member renaming (e.g. `InterfaceName` → `interface_name` in Scriban). The rendered `.md` files are **not committed to git** — they are generated at build time by `RegenerateSkillDocs` (runs after build) and embedded as resources. Only edit the `.md.scriban` templates in `SkillTemplates/`. Skills without template variables are also stored as `.md.scriban` and pass through Scriban unchanged.

### Installing skills for non-MCP agents

`beam install-ai-skill` extracts embedded skill files to `.claude/skills/`, `.cursor/skills/`, `.windsurf/skills/`, or `.opencode/skills/`. Supports `--force` to overwrite and auto-detects agent directories.

### Available skills

| Skill | Purpose |
|---|---|
| `beam-initialize-project` | Environment selection (prod/staging/dev/custom), CID/PID concepts, `init` with `--host`, verify config |
| `beam-manage-service` | Create, configure, and remove microservices with storage, federation, and local dev patterns |
| `beam-create-portal-extension` | Discover mount options, create page or component extension, add microservice dependencies, run, open in browser |
| `beam-build-and-deploy` | Build Docker images with `deploy plan`, release with `deploy release` |
| `beam-diagnose-project` | Systematic diagnostic checklist for workspace, auth, build, Docker, and network issues |
| `beam-login-auth` | Username/password and refresh token auth, token persistence |
| `beam-content-types` | Create and manage content types |
| `beam-content-workflow` | Content sync and publish workflows |
| `beam-content-history` | Content change history |
| `beam-get-source` | Read Beamable SDK source code — file content with pagination for sandboxed environments |
| `beam-openapi` | OpenAPI spec generation |
| `beam-review` | Code review checklist for Beamable projects |
| `beam-unity-examples` | Unity getting started and sample projects — leaderboards, matchmaking, multiplayer, IAP, microservices |
| `beam-unreal-guide` | Unreal development guide — custom content types, microservice integration, Blueprints, semantic types |
| `beam-microservice-examples` | C# microservice quick-start — endpoints, storage, API patterns, server-authoritative example |
| `beam-web-guide` | Web SDK guide — installation, authentication, stats, leaderboards, content, announcements, accounts, and microservice clients |

### Adding new skills

1. Create `.md.scriban` template in `cli/Docs/SkillTemplates/` with `name` and `description` frontmatter
2. Build the project — `RegenerateSkillDocs` renders templates into `cli/Docs/Skills/` automatically
3. The rendered `.md` is auto-embedded by the `<EmbeddedResource Include="Docs\**\*.md" />` glob
4. Do **not** commit the generated `Skills/*.md` files — only the `.md.scriban` sources are tracked

---

## How to Use

```bash
dotnet tool install -g Beamable.Tools   # or build from source: ./dev.sh
beam mcp setup                          # writes .mcp.json to project dir
```

Reload your AI client — it discovers `.mcp.json` and exposes 5 `beam_*` tools. Both `beam mcp setup` and `beam mcp serve` implement `IStandaloneCommand` — they work without a `.beamable` config directory.

---

## mcpb Distribution

The `.mcpb` format is a bundle for Claude Desktop Extensions. It packages a self-contained `beam` CLI binary with the MCP server configuration so users can install the Beamable extension without needing a .NET SDK or global tool install.

### Build process

The `build-mcpb.sh` / `build-mcpb.ps1` scripts in `cli/mcpb/` build bundles for 4 platform runtime identifiers:

1. **Publish the CLI binary**: `dotnet publish cli/cli.csproj -f net8.0 -r {RID} -c Release --self-contained true -p:PublishSingleFile=true`
2. **Rename binary**: `Beamable.Tools.exe` → `beam.exe` (Windows) or `Beamable.Tools` → `beam` (Unix)
3. **Template manifest**: Replace `VERSION_PLACEHOLDER` and `BINARY_PLACEHOLDER` in `manifest.json`
4. **Pack bundle**: `mcpb pack $STAGE output/beamable-{RID}.mcpb`

**Note on `VERSION_PLACEHOLDER`:** Both `manifest.json` (mcpb bundle manifest) and `server.json` (MCP Registry metadata) contain `VERSION_PLACEHOLDER`. The `build-mcpb.sh` script only replaces it in `manifest.json` during the mcpb build. For MCP Registry publishing, `server.json` must be updated separately — see "Publishing to MCP Registries" below.

### Output bundles

| File | Platform |
|---|---|
| `beamable-win-x64.mcpb` | Windows (x64) |
| `beamable-osx-arm64.mcpb` | macOS (Apple Silicon) |
| `beamable-osx-x64.mcpb` | macOS (Intel) |
| `beamable-linux-x64.mcpb` | Linux (x64) |

### Release

Bundles are uploaded to GitHub Releases with tag `cli-{VERSION}`. Building is gated on the `buildMcpb` workflow_dispatch input (default `false`) and only runs for production releases when `dryRun == false`.

### Installation paths

1. **Claude Desktop Extensions directory**: Search for "Beamable" and install directly
2. **Manual download**: Download the `.mcpb` for your platform from GitHub Releases, then open it in Claude Desktop

### Difference from `beam mcp setup`

| | mcpb | `beam mcp setup` |
|---|---|---|
| Requires .NET SDK | No (self-contained binary) | Yes |
| Installation | Claude Desktop Extensions or `.mcpb` file open | Writes `.mcp.json` to project dir |
| CLI version | Bundled with the mcpb | Project-local via `.config/dotnet-tools.json` |
| Best for | Claude Desktop / Cowork | Cursor, Windsurf, VS Code, other MCP clients |

---

## Release Workflow

mcpb bundles are built and uploaded as part of the `Release Nuget Packages (and CLI)` workflow (`.github/workflows/release-nuget.yml`). The workflow is triggered via `workflow_dispatch` with these relevant inputs:

| Input | Type | Description |
|---|---|---|
| `releaseType` | choice | `nightly`, `rc`, `exp`, or `production` |
| `dryRun` | boolean | When true, no publish happens |
| `buildMcpb` | boolean | Build and upload mcpb bundles (default `false`) |

The 3 mcpb steps are gated on `inputs.buildMcpb && inputs.releaseType == 'production' && inputs.dryRun == false`:

1. **Install mcpb CLI**: `npm install -g @anthropic-ai/mcpb`
2. **Build mcpb bundles**: Runs `bash build-mcpb.sh` in `cli/mcpb/` with `VERSION` env var set from the release version step
3. **Upload mcpb bundles to release**: Iterates `cli/mcpb/output/*.mcpb` and uploads each to the GitHub Release tag `cli-{VERSION}` via `gh release upload --clobber`

To trigger a production release with mcpb bundles: go to Actions → "Release Nuget Packages (and CLI)" → Run workflow, set `releaseType: production`, `dryRun: false`, `buildMcpb: true`.

---

## Publishing to MCP Registries

Step-by-step instructions for registering the Beamable MCP server with external registries so users can discover and install it.

| Registry | Format | Submission | Target |
|---|---|---|---|
| Official MCP Registry | NuGet package | `mcp-publisher` CLI | Code agents (Cursor, VS Code, Windsurf) |
| Smithery | mcpb bundle | `mcpb` CLI | Commercial marketplace |
| Claude Desktop Extensions | mcpb bundle | Anthropic partner program | Sandboxed agents (Claude Desktop, Cowork) |
| Glama | GitHub repo URL | Web submission (auto-indexed) | General MCP discovery |
| PulseMCP | GitHub repo URL | Web form (reviewed) | Curated MCP discovery |
| mcp.so | GitHub repo URL | Web / GitHub issue | Community MCP discovery |
| awesome-mcp-servers | GitHub repo URL | Pull request | Developer reference list |

### Official MCP Registry (`registry.modelcontextprotocol.io`)

The official MCP Registry is a metaregistry maintained by the MCP project. It indexes server metadata from a `server.json` file in the repository. Beamable already has this file at `cli/mcpb/server.json` (schema `2025-12-11`, name `io.github.beamable/beam-mcp`, NuGet package `Beamable.Tools`).

**Note:** `server.json` is the MCP Registry metadata file — it describes the MCP server as a NuGet package. It is separate from `manifest.json`, which is the mcpb bundle manifest for Claude Desktop Extensions. Both live in `cli/mcpb/` but serve different registries.

**Steps:**

1. **Install the publisher CLI:**
   ```bash
   npm install -g mcp-publisher
   ```

2. **Authenticate via GitHub OAuth:**
   ```bash
   mcp-publisher auth login
   ```
   This opens a browser for GitHub OAuth. The authenticated GitHub user must have write access to the `beamable/BeamableProduct` repository.

3. **Set the version in `server.json`:**
   The file contains `VERSION_PLACEHOLDER` in the `version` and `packages[0].version` fields. Replace both with the actual release version before publishing:
   ```bash
   sed -i 's/VERSION_PLACEHOLDER/1.2.3/g' cli/mcpb/server.json
   ```
   Use the version from your latest `cli-{VERSION}` GitHub Release tag. **Do not commit the version change** — the placeholder must remain in source for the next release.

4. **Validate the server.json:**
   ```bash
   mcp-publisher validate cli/mcpb/server.json
   ```
   Fix any schema errors before publishing.

5. **Publish:**
   ```bash
   mcp-publisher publish cli/mcpb/server.json
   ```
   The registry verifies:
   - The `name` field (`io.github.beamable/beam-mcp`) maps to the `repository.url` GitHub org
   - The authenticated user has write access to that repository
   - The NuGet package `Beamable.Tools` exists on nuget.org

6. **Verify listing:** Visit `https://registry.modelcontextprotocol.io/servers/io.github.beamable/beam-mcp` after publishing.

**Automating via GitHub Actions (optional):** The `mcp-publisher` CLI supports GitHub OIDC tokens. Add a step to the release workflow:
```yaml
- name: Publish to MCP Registry
  if: ${{ inputs.releaseType == 'production' && inputs.dryRun == false }}
  run: |
    npm install -g mcp-publisher
    sed -i 's/VERSION_PLACEHOLDER/${{ steps.version.outputs.VERSION }}/g' cli/mcpb/server.json
    mcp-publisher publish cli/mcpb/server.json
  env:
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

### Smithery (`smithery.ai`)

Smithery is a commercial MCP marketplace with a web directory and CLI for publishing.

**Steps:**

1. **Create a Smithery account:** Sign up at `https://smithery.ai` and link your GitHub account.

2. **Install the Smithery CLI:**
   ```bash
   npm install -g @anthropic-ai/mcpb
   ```
   (The `mcpb` CLI handles Smithery publishing — same tool used for building bundles.)

3. **Authenticate:**
   ```bash
   mcpb auth login
   ```

4. **Publish the mcpb bundle:**
   ```bash
   mcpb publish output/beamable-{RID}.mcpb
   ```
   Publish each platform bundle separately, or publish a single bundle and Smithery will list it for the matching platform.

5. **Verify listing:** Search for "Beamable" on `https://smithery.ai` after publishing.

**Note:** Smithery may require additional metadata fields in `manifest.json` as their schema evolves. Check `mcpb publish --help` for current requirements.

### Claude Desktop Extensions Directory

There is **no public self-service submission process** for the Claude Desktop Extensions directory. The directory is curated by Anthropic through partner relationships.

**Options:**
- **Contact Anthropic partnerships** to request inclusion — reach out via your existing Anthropic contacts or the partner form at `https://anthropic.com/partners`
- **GitHub Releases + manual install** — users download the `.mcpb` from GitHub Releases and open it in Claude Desktop (this already works today)
- **MCP Registry listing** — once published to the official MCP Registry, Claude Desktop may index it automatically in future updates (the registry is designed as a metaregistry that clients can pull from)

### Glama (`glama.ai/mcp/servers`)

Glama is a premium MCP registry with 23,000+ servers. It auto-indexes from GitHub repositories daily.

**Steps:**
1. Go to `https://glama.ai/mcp/servers` and click "Submit Server"
2. Enter the GitHub repository URL: `https://github.com/beamable/BeamableProduct`
3. Glama auto-indexes the repo and refreshes daily — no CLI or manual updates needed after initial submission

### PulseMCP (`pulsemcp.com`)

PulseMCP is a hand-reviewed MCP directory with 14,000+ servers, maintained by members of the MCP Steering Committee. Submissions go through a curated review process.

**Steps:**
1. Go to `https://www.pulsemcp.com/submit`
2. Submit the GitHub repository URL and a brief description of the MCP server
3. Wait for review — PulseMCP editors verify the server works and add metadata (security analysis, category tags)

### mcp.so

mcp.so is the largest community-driven MCP directory with 20,000+ servers.

**Steps:**
1. Go to `https://mcp.so` and click "Submit"
2. Provide the GitHub repository URL and server description
3. The listing appears after basic validation — no editorial review

### awesome-mcp-servers (GitHub)

The `awesome-mcp-servers` repository is the de facto standard curated list of MCP servers. Many other directories (including mcpservers.org) mirror it.

**Steps:**
1. Fork `https://github.com/punkpeye/awesome-mcp-servers`
2. Add Beamable to the appropriate category in `README.md` following the existing format
3. Open a pull request — maintainers review and merge
4. Other directories that mirror this list will pick up the entry automatically

---

## Manual Testing

```bash
npx @modelcontextprotocol/inspector beam mcp serve
```

Opens a browser UI at `http://localhost:5173` to call each tool and inspect raw JSON responses. `CliException` errors are captured from stderr and returned as `{"error": "..."}` JSON lines.


.github/workflows/release-nuget.yml
```diff
@@ -220,3 +225,23 @@ jobs:
...
+        env:
+          VERSION: ${{ steps.version.outputs.VERSION }}
+
+      - name: Upload mcpb bundles to release
+        if: ${{ inputs.buildMcpb && inputs.releaseType == 'production' && inputs.dryRun == false }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          for f in cli/mcpb/output/*.mcpb; do
+            gh release upload "cli-${{ steps.version.outputs.VERSION }}" "$f" --clobber
```
cdhanna, line 246: so the mcpb file gets associated to our github release?

microservice/microservice/microservice.csproj
```diff
@@ -63,5 +63,9 @@
   <ItemGroup Label="Pack Internal Beamable References">
     <None Include="Targets/Beamable.Microservice.Runtime.*" PackagePath="build" Pack="True"/>
   </ItemGroup>
-  
+
+  <ItemGroup Label="Pack: source code">
+    <None Include="**/*.cs" PackagePath="content/net8.0" Pack="True" Exclude="bin/**;obj/**" />
```
cdhanna: we are packing the source into net8; but what about when building the net10 part? I guess this is "fine", but it feels like a strange convention to set in stone. Same for the tooling.common src packing as well.

cli/cli/App.cs
```diff
@@ -1522,12 +1543,68 @@ static void PrintOutputHelp(HelpContext context)
...
+	private static void WarnIfAIEnvironmentWithoutMcp(string[] args)
+	{
+		if (IsRunningInMcpServer) return;
+		var joined = string.Join(" ", args).ToLowerInvariant();
+		if (joined.Contains("mcp serve") || joined.Contains("mcp setup")) return;
+
+		var aiAgent = DetectAiAgent();
+		if (aiAgent == null) return;
+
+		Console.Error.WriteLine(
```
cdhanna: this is using `Console` directly- but at this point, I think we have pretty conventional logging practices, right? I worry that using `Console` directly may not respect all of our random/various logging conventions.

.github/workflows/release-nuget.yml
DiasAtBeamable: Yes, that's the idea, so we can share it with any agentic tools that support MCPB. But for Claude Code, the idea is to make it available from their extension list.

cli/cli/App.cs
```diff
@@ -1522,12 +1543,68 @@ static void PrintOutputHelp(HelpContext context)
...
+			"[beam] You are calling beam CLI directly. For better AI integration, use the Beamable MCP server. " +
+			"Run 'beam mcp setup' to generate a .mcp.json config, then use MCP tools (beam_exec, beam_get_help, beam_get_skill) for structured interaction.");
+	}
+
+	internal static string DetectAiAgent()
```
cdhanna: splitting hairs in an AI world- but I would rather this method return a `bool` with an out parameter for the agent name. Right now, the `null` return is a special type of value, and I'd like that to be more explicit. 

Also, no need to make this `internal`.
Something like,

```csharp
public static bool TryDetectAiAgent(out string agentName)
```

microservice/microservice/microservice.csproj
DiasAtBeamable: So, it was a constraint that I set, thinking that .NET 8.0 would be the most commonly used, but yeah, it might not be true...

cli/cli/Commands/EnvCommand.cs
```diff
@@ -0,0 +1,63 @@
...
+		var aiDetection = new Dictionary<string, string>();
+		foreach (var name in aiVarNames)
+		{
+			var value = Environment.GetEnvironmentVariable(name);
+			aiDetection[name] = value ?? "(not set)";
+		}
+
+		return Task.FromResult(new EnvCommandResult
+		{
+			environmentVariables = envDict.OrderBy(kv => kv.Key).ToDictionary(kv => kv.Key, kv => kv.Value),
```
cdhanna: what is the purpose of having a command that dumps the env? an agent could just run `env`, right? This feels a bit too much, but open to hearing otherwise.

cli/cli/App.cs
DiasAtBeamable: So, for this case, I wanted to not use our logging so that in case this could be ignored, as in the first pass it was  Debug, and later I changed to Error. Make sense to use our logging conventions now that it is an error.

cli/cli/Commands/Mcp/McpToolExecutor.cs
```diff
@@ -0,0 +1,693 @@
...
+					break;
+				}
+				case "cli":
+				{
+					var nugetBase = Path.Combine(
+						Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
+						".nuget", "packages");
+					var commonSrc = Path.Combine(nugetBase, "beamable.common", detectedVersion, "content", "netstandard2.0", "Runtime");
+					var toolingSrc = Path.Combine(nugetBase, "beamable.tooling.common", detectedVersion, "content", "netstandard2.1");
+					var runtimeSrc = Path.Combine(nugetBase, "beamable.microservice.runtime", detectedVersion, "content", "net8.0");
```
cdhanna: yep- here is the other side of the hardcoded net8. 

I feel like we should just pick a different conventional spot in our package to host the src code.

cli/cli/Commands/Mcp/McpGroupCommand.cs
```diff
@@ -0,0 +1,8 @@
+namespace cli.Mcp;
+
+public class McpGroupCommand : CommandGroup, IStandaloneCommand
```
cdhanna, line 3: mark this command as ForInternalUse or whatever, for now.

cli/cli/Commands/EnvCommand.cs
DiasAtBeamable: Ops, this one shouldn't exist, I forgot to remove it. I created it to test that the data was correct, because the AI agents were not returning good info when I asked for env. So I created this command to be called using "!" in Claude and opencode.

DiasAtBeamable: I'll remove it.

cli/cli/Commands/Mcp/McpToolExecutor.cs
DiasAtBeamable: Yeah, make sense, I'll search if it is possible to put it directly in the content folder.

cli/cli/Commands/Mcp/McpServerBuilder.cs
```diff
@@ -0,0 +1,165 @@
...
+				Directory.SetCurrentDirectory(uri.LocalPath);
+				return;
+			}
+		}
+		catch
+		{
+			// Client doesn't support roots — fall through to env var
+		}
+
+		var workspace = Environment.GetEnvironmentVariable("BEAM_WORKSPACE");
```
cdhanna: is this a new or existing beam env variable? I don't think I recognize this name

DiasAtBeamable: Another thing that I forgot to remove, the MCPB has one option to add a workspace for it to work, but while testing, I found that it was not what I thought that it would be, it is a workspace for Claude to use, not per project/session. 
I'll remove it. Thanks for noticing this.

microservice/microservice/microservice.csproj
DiasAtBeamable: Done

cli/cli/Commands/EnvCommand.cs
DiasAtBeamable: Done

cli/cli/Commands/Mcp/McpGroupCommand.cs
DiasAtBeamable: Done

cli/cli/Commands/Mcp/McpServerBuilder.cs
DiasAtBeamable: Done

cli/cli/Commands/Mcp/McpToolExecutor.cs
DiasAtBeamable: Done, now they are in the content/sourceCode folder.

cli/cli/App.cs
DiasAtBeamable: Done

PR #4593, opened 2026-04-30
Fix serialization issue with Portal Extensions data

gabrielbeamable:
The deserialization of the `metadata.json` was not catching the actual Label and Group fields, and instead it was sending the full type name to Portal to render.

I decided to fix this in the Portal Extension data types level because I'm not sure of repercussions on changes in the OptionalString directly, since it's shared with multiple projects.


cli/cli/Services/BeamoLocalSystem_PortalExtension.cs
```diff
@@ -264,15 +265,42 @@ public class PortalExtensionMountProperties
...
+	[JsonConverter(typeof(OptionalStringConverter))]
+	public OptionalString NavIcon;
+
 	[JsonProperty(KEY_NAV_GROUP_ORDER)] public OptionalInt NavGroupOrder;
 	[JsonProperty(KEY_NAV_LABEL_ORDER)] public OptionalInt NavLabelOrder;
 
 	[JsonProperty("args")] public Dictionary<string, string> Args = new Dictionary<string, string>();
 }
 
+public class OptionalStringConverter : JsonConverter<OptionalString>
```
gabrielbeamable: we already have a converter for this, find and use that

PR #4598, opened 2026-05-14
Add omitTags option to manifest requests to control tag emission

### LINKED ISSUE #4597: Implement Content tags opt-out with omitTags=false in manifest requests
Some customers heavily use tags on the backend for content organization, but don't use them on the client at all. Including a lot of tags with each entry in the manifest bloats the manifest response, and it would be useful, in specific circumstances, to opt-out from fetching content tags on the client.

dpatekar:
# Ticket

https://github.com/beamable/BeamableProduct/issues/4597

# Brief Description

Adds opt-in `OmitContentManifestTags` toggle to `ContentConfiguration`. When enabled, the SDK appends `omitTags=true` to the manifest fetch URL (`/basic/content/manifest/public?id=...&omitTags=true`), asking the backend to emit an empty tags column/array for every entry. Default is false — existing projects send the exact same request they do today.

  - New Inspector field on `ContentConfiguration` with a tooltip that calls out the trade-off.
  - `ContentParameterProvider` and `ManifestSubscription` plumb the flag from config to URL.
  - Existing pattern matched: same shape as `EnableLocalContentInEditor`.

  Useful for customers (e.g. the BPC realm: ~6.5× manifest growth, 72KB → 475KB) that use content tags purely for server-side organization and don't read them on the client. Tag-based `ContentQuery` filters such as `"tag:weapon"` return nothing when enabled — game devs that opt in must not depend on them.

  Resolves BEAM-4597. Requires server-side `omitTags` support: BeamableBackend#814.

# Checklist


cdhanna:
Copilot's suggestion is not important here. The param is optional, so it isn't going to break anyone.

PR #4601, opened 2026-05-14
Fix - Editor freeze when telemetry collector ps is watching

### LINKED ISSUE #4589: Editor freeze when telemetry collector ps is watching
## Summary

The Unity Editor freezes briefly (noticeable pause) approximately every 15 seconds while the Beamable SDK is loaded. The freeze coincides with the `telemetry collector ps --watch=True` CLI process delivering its periodic `extraStream` payload back to the editor.

## Stack trace

```
CLI: received, {"ts":...,"type":"extraStream","data":{"FileCount":0,"FolderSize":0,"LastPublishTimestamp":0}}, from ... telemetry collector ps --watch=True
UnityEngine.Debug:Log (object)
Beamable.Common.Spew.Logger:DoSimpleSpew (string)
Beamable.Editor.BeamCli.CliLogger:Log (object[])
Beamable.Editor.BeamCli.BeamWebCommand/<>c__DisplayClass10_0:<Run>b__0 ()
Beamable.Editor.BeamableDispatcher/<Scheduler>d__12:MoveNext ()
Unity.EditorCoroutines.Editor.EditorCoroutine:ProcessIEnumeratorRecursive
Unity.EditorCoroutines.Editor.EditorCoroutine:MoveNext ()
UnityEditor.EditorApplication:Internal_CallUpdateFunctions ()
```

## Observed behavior

- A brief but noticeable editor freeze occurs approximately every 15 seconds, both in edit mode and on entering play mode.
- The freeze also occurs in a fresh empty project with Beamable installed but without signing in.
- Cancelling the `telemetry collector ps` command in the Beamable CLI Debugger (Beamable > Utilities > CLI Debugger) stops the freeze, but the command restarts on every domain reload (script compile, enter play mode, etc.).

## Expected behavior

The periodic CLI subprocess response should not cause a perceptible stall on the Unity main thread. The `BeamableDispatcher` callback handling the `extraStream` result (`BeamWebCommand.cs:453`) should either be non-blocking or the work should be dispatched off the main thread.

## Environment

- Unity SDK: 5.0.0
- Unity: 6000.3.8f1
- OS: not recorded

DiasAtBeamable:
# Ticket

https://github.com/beamable/BeamableProduct/issues/4589

# Brief Description

- Increased OtelCollectorStatus Delay from 15s to 60s. It was causing GC overload on Unity. As with 60 seconds, it is not causing GC overload anymore.
- Fixed CalculateDirectorySize to use EnumerateFiles than using a Recursive call.
- Update BeamWeb read task to use TaskCompletionSource rather than a Task.Delay.


cdhanna (approved):
this looks good-
I am sort of curious why the EnumerateFiles one is better than the previous version.

Also, not scheduling the tasks in the beamWebCommand is _also_ interesting. I think it is fine to try this out in our QA war, but in general, I am _slightly_ worried that we could get some strange thread access violation things here. In general, the tasks can execute on non-main threads, and if any of the web command invocations get continued off the main thread, but try to acces unity main thread state; KABOOM. 

So... fine, with a "we need to keep an eye on this."

Maybe we end up moving certain continuations into the scheduler, rather than ALL of them.

client/Packages/com.beamable/CHANGELOG.md
```diff
@@ -9,10 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Deserialization issue with `properties` field in Score Items of Events
 - Fixed an issue where the Unity Editor would not detect changes to Icon subObject (for Sprites in Multiple Mode) and thus not saving it correctly
-<<<<<<< fix/contentNullFields
 - Fixed an issue when creating a new Content Object some Optional and string values were null instead of default values.
-=======
 - Fix issue where Critical log errors where not being parsed to Unity LogLevels.
+- Fix periodic freeze caused by telemetry collector polling.
 >>>>>>> main
```
allister-beamable: This looks like a leftover conflict marker, and should be removed.

DiasAtBeamable:
> I am sort of curious why the EnumerateFiles one is better than the previous version.
@cdhanna 

EnumerateFiles begins yielding files as they are discovered, while GetDirectories must finish the entire search before starting the loop. So we gain some extra performance ms here, and there, it will always depend on how many OTEL Logs and traces are there to find.

DiasAtBeamable:
> Also, not scheduling the tasks in the beamWebCommand is _also_ interesting. I think it is fine to try this out in our QA war, but in general, I am _slightly_ worried that we could get some strange thread access violation things here. In general, the tasks can execute on non-main threads, and if any of the web command invocations get continued off the main thread, but try to acces unity main thread state; KABOOM.
@cdhanna 

It is still being scheduled. I just joined the two schedules: the CliLogger with the process task. I think, since it is only a log, it is okay to let it be scheduled with the tasks. Do you think that's okay?

allister-beamable (approved):
Looks reasonable! I think Chris's threading concerns are valid, but it does seem like the only truly effective way to test them is "in the wild". Furthermore, I expect that even if there are concurrency issues, they will not be any worse for UX than what came before.

Ship it!

gabrielbeamable (approved):
looks good overall, 
but I got to be honest, I dont understand what is going on with the changes in the `BeamWebCommand.cs` file. 

What is the difference between using the `readTask.IsCompleted` and using a delay, and passing a `TaskCompletionSource` to the `Task.WhenAny`? What does this fix?

Not blocking the merge of this, just curious to know how this stuff works and why it was problematic.

DiasAtBeamable:
> What is the difference between using the `readTask.IsCompleted` and using a delay, and passing a `TaskCompletionSource` to the `Task.WhenAny`? What does this fix?

@gabrielbeamable 

The old code was using a while read with a Task.Delay until it finishes. As the new one only needs to await the TaskCompletionSource to finish.

PR #4602, opened 2026-05-14
Fix for signed request failing when running local microservice with otel enabled

gabrielbeamable:
When we call `project generate-env` running the microservice locally, we don't pass the flag to generate the secret and without the secret, the signed request won't work. This workaround is here to make it use a refresh token instead of the secret. I'm only choosing this approach because I believe that there might be a good reason to not generate the secret locally.


cdhanna (approved):
seems fine; nice

allister-beamable (approved):
Looks reasonable. @gabrielbeamable can you explain for my uninitiated brain what caused the problem this solves? In production, are the socket-y things that get filtered out already set to `http://` or `https://`?

allister-beamable:
> Looks reasonable. @gabrielbeamable can you explain for my uninitiated brain what caused the problem this solves? In production, are the socket-y things that get filtered out already set to `http://` or `https://`?

For those watching from home, the explanation is that using OTel locally is an unusual workflow that does not really get exercised much, but the configuration for cloud-deployed services using OTel is fine -- it just didn't work nicely with local services before this patch.

PR #4606, opened 2026-05-19
Adds visible Content Manager progress during realm sync/revert operations and improves content download resilience for large syncs

### LINKED ISSUE #4605: SSL/socket failures when syncing high number of contents
## Describe the Bug 

SSL/socket failures when syncing high number of contents, usually happening above 300 contents. 

## To Reproduce
- delete local content
- In a realm with hundreds of content, go to beam content manager
- sync content

### Expected Behavior

All contents downloaded without any issues

### Actual Behavior

SSL failures in the cli debugger during content sync

## Metadata

- SDK version: 5.0.1
- Engine version: Unity 6+
- Operating System: Windows 11

## Additional Context

Known workaround: keep trying

mhijaziB:
# Ticket

Fixes #4605

# Brief Description

Details

Added an in-window Content Manager progress display for sync/revert flows.

Shows operation title, processed item count, total item count, progress bar fill, and current content item name.
Keeps sync progress inside the Content Manager instead of relying on Unity’s modal progress UI.
Disables revert/action buttons while a sync operation is active to avoid overlapping operations.
Uses delayed window-state transitions to avoid IMGUI layout mismatches during repaint.
Improved CLI content download behavior during content sync.

Added a dedicated content-file download path in ContentService.
Uses a shared HttpClient for content file downloads instead of routing every CDN file through the generic verbose CLI requester.
Adds transient retry handling with jittered backoff for socket resets, timeouts, and retryable HTTP responses.
Removes per-file verbose “Downloading content with id...” sync log noise from the Unity CLI Debugger.
Keeps large syncs faster and cleaner while avoiding the SSL/socket reset failures seen during high-content-count tests.
Added sync download concurrency control.

New CLI flag: --download-max-parallel-count
New Unity content configuration field: EditorSyncDownloadMaxParallelCount
The editor passes this value into content sync when configured.
CLI default is 64.
Unity unset means the CLI default is used.
0 means unbounded parallelism.

Tests:
- stress testing with 2k+ content amount
- confirmed consistently that no SSL/socket reset failures are happening during content sync

# Checklist

### DIFF
--- a/cli/cli/Services/Content/ContentService.cs
+++ b/cli/cli/Services/Content/ContentService.cs
@@ -1253,8 +1253,14 @@ public partial class ContentService
 	}
 
 	/// <summary>
-	/// Determines whether a content download failure is likely transient and should be retried.
+	/// Determines whether a content download failure is likely caused by temporary transport or service pressure
+	/// and should be retried instead of failing the entire sync operation.
 	/// </summary>
+	/// <remarks>
+	/// Content sync can download hundreds or thousands of small files in parallel. In that scenario, occasional
+	/// TLS/socket resets, timeouts, rate limits, or temporary server errors can happen even when the content is valid.
+	/// Retrying only those transient failures prevents one short-lived network issue from aborting the full sync.
+	/// </remarks>
 	private static bool IsTransientContentDownloadException(Exception exception)
 	{
 		if (exception is RequesterException requesterException)
--- a/cli/cli/Services/Content/ContentService.cs
+++ b/cli/cli/Services/Content/ContentService.cs
@@ -78,6 +78,13 @@ public partial class ContentService
 	/// </summary>
 	private const string FAKE_EMPTY_MANIFEST_UID = "EmptyManifest";
 
+	private const int CONTENT_DOWNLOAD_MAX_ATTEMPTS = 4;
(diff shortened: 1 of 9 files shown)


cli/cli/Services/Content/ContentService.cs
```diff
@@ -1163,6 +1192,89 @@ public async Task<ContentSyncReport> SyncLocalContent(ClientManifestJsonResponse
...
+		return TimeSpan.FromMilliseconds(exponentialDelayMs + jitterMs);
+	}
+
+	/// <summary>
+	/// Determines whether a content download failure is likely transient and should be retried.
+	/// </summary>
+	private static bool IsTransientContentDownloadException(Exception exception)
+	{
+		if (exception is RequesterException requesterException)
+		{
```
DiasAtBeamable, line 1267: Could you add a comment here or in the summary, why do we need to check for transient errors.

DiasAtBeamable (approved):
Seems good to me, I just added a suggestion to add a comment so we know why we need to check for transient error downloads in the future when changing this code.

allister-beamable (approved):
Uses more ternary operator than I would usually prefer, and I never heard of `Mathf.Clamp01` before today, but overall looks good!

mhijaziB:
@allister-beamable The `Mathf.Clamp01` is more of a Unity C# only function

cdhanna (approved):

PR #4630, opened 2026-05-28
Fix bug when Adding Sample to the Scene and Playing quickly

VitorBeamable:
Try to fix it forcing a AssetDatabase.Refresh() and cleanup unused enum


cdhanna (commented):
how does this fix the issue? I think now an asset refresh would happen even if the scene was already copied locally; which probably doesn't need to be true?

PR #4634, opened 2026-05-29
Feature - Add rename tracking and UI support in content service (Unity Only)

DiasAtBeamable:


allister-beamable:
@DiasAtBeamable looks like it accidentally includes Stripe-related stuff?

allister-beamable (approved):
I did not dig into the details too much, but the surface looks fine. No nasty recursion, reasonable use of loops, blends with pre-existing behavior reasonably well.

Slight concern that the fix may be heavier than the problem merits, but I am willing to defer to your judgment @DiasAtBeamable as well as @cdhanna on that.

PR #4643, opened 2026-06-05
Content manifest caching #4435

### LINKED ISSUE #4435: Content Manifest Caching
## Reason for the Feature

Once content is downloaded it is stored on device so later on only the new or changed content needs to be downloaded on next run. We are not doing that for manifest, you need to download latest manifest on each run. 

We can do better. We can get the UID of the current manifest for given namespace and we have it already downloaded we can skip to the next step instead of downloading whole manifest. To do so we need to start caching Manifests just like we do for Content items. It would require changing the `GetCurrent` method in ContentService to instead of downloading latest manifest do two steps:
- call `/basic/content/manifest/checksum?id={ManifestID}` to get information about the uid of the current manifest.
- if that is not stored on the device download it with `/basic/content/manifest/public?id={ManifestID}&uid={ManifestUID}` and then store on device so it will available on next run of the game.

It should really speed up subsequent startup times of the SDK. It would be also useful to use the baked manifest so if there would be no change between baking and running the game the only call would be for getting the latest manifest UID and confirming that it is the current one.

Leinnan:
Resolves #4435 .

In my test when I was using the content manifest with less than 100 entries the time to get the content manifest were similar. When I've increased the manifest size to have bigger size (~1600 entries) the difference start to grow, but it was not consistent. I would say that the biggest win would be for cases where there is a big content manifest that changes from time to time so the clients would need to download a lot of data less frequently. There could be cases where there is a small manifest that does change often, then it could be worse than current solution. Instead of current one request(get me the latest manifest) it will perform two (get me the latest manifest UID, then the manifest with that UID) which is why I've introduced option to opt out of it with define symbol `BEAMABLE_OPTOUT_MANIFEST_FILE_CACHING`.


allister-beamable (changes requested):
Generally looks fine, and should be a net improvement (or at least break even) for most usage patterns. Two things:
- needs a changelog entry
- is there any "safety valve" for the (hopefully rare) case of a corrupted or partially-incomplete cached manifest? some client-side way to force it to retry?

Leinnan:
> Generally looks fine, and should be a net improvement (or at least break even) for most usage patterns. Two things:
> 
>     * needs a changelog entry
> 
>     * is there any "safety valve" for the (hopefully rare) case of a corrupted or partially-incomplete cached manifest? some client-side way to force it to retry?

Changelog was updated, about safety valve I do not see a easy way to add it now as part of this work.

cli/cli/CHANGELOG.md
```diff
@@ -5,7 +5,24 @@ All notable changes to this project will be documented in this file.
...
+### Fixed
+- Portal extension scanning no longer excludes sym linked package files
+
+## [7.2.2] - 2026-07-16
+
+### Fixed
+
+- Fixed an issue where SmallerJson was not serializing/deserializing null fields to default values.
+
+## [7.2.1] - 2026-06-30
```
allister-beamable, line 20: Good call, updating the log to accurately reflect the release date! (and a good reminder to the whole team to update it when cutting releases -- I think this specific one is my fault 😇)

allister-beamable (approved):
Looking good with the CHANGELOG updates.

Having discussed & thought about the "safety valve" topic: hopefully the circumstances that could lead to an actually corrupted or incomplete cache are vanishingly rare, and if they do occur the solution would be to uninstall & reinstall or clear device cache. Honestly, any mechanism trying to forestall or react to corrupted/incomplete caches would be more complex and error prone than the simple "turn it off and on again" hammer of uninstall & reinstall.

PR #4647, opened 2026-06-08
Generate portal extension microservice clients on build

gabrielbeamable:
## Summary
  - Add a new CLI command `beam project generate portal-extension-clients` (alias `pe-client`) that generates TypeScript API clients for portal extensions from their dependent microservices' OpenAPI specs
  - Extract OpenAPI document parsing into a reusable `TryGetOpenApiDocument` static method on `BeamoServiceDefinition`, shared by both the new command and existing federation data loading
  - Wire client generation into the microservice build pipeline via an MSBuild target (`GeneratePortalExtensionClients`) that runs after each microservice build
  - Remove the old on-file-change client generation call from `PortalExtensionObserver` in favor of the build-time approach

  ## Changes

  ### New CLI Command (`GeneratePortalExtensionClientsCommand`)
  - Registered under `project generate portal-extension-clients` with alias `pe-client`
  - Accepts an optional `--service-name` flag; when omitted, generates clients for all HTTP microservices
  - For each microservice, finds portal extensions that depend on it, reads the microservice's OpenAPI document, and generates TypeScript client code into `<extension>/beamable/clients/`
  - Uses `Parallel.ForEachAsync` with a concurrency limit of 8 and path-level locking to safely handle multiple microservices writing to the same extension directory

  ### Refactored OpenAPI Parsing (`BeamoLocalSystem.cs`)
  - Extracted `TryGetOpenApiDocument(string openApiPath)` as a public static method returning `(bool, OpenApiDocument)`
  - `ReloadFederationsData` now delegates to this method instead of inlining the parsing logic

  ### Build-Time Integration (`Beamable.Microservice.Runtime.targets`)
  - New MSBuild target `GeneratePortalExtensionClients` runs after `AfterBuild`, depends on `generate-oapi`
  - Invokes `dotnet beam project generate pe-client --service-name $(BeamId)` so clients are regenerated whenever a microservice is built

  ### Removed File-Watch Trigger (`PortalExtensionDiscoveryService.cs`)
  - Removed the call to `PortalExtensionAddDependencyCommand.GenerateDependenciesClients` from the file watcher observer, since client generation now happens at build time

  ## Test plan
  - [ ] Build a microservice that has a dependent portal extension and verify `.ts` client files are generated in `<extension>/beamable/clients/`
  - [ ] Run `dotnet beam project generate pe-client` without `--service-name` and confirm it generates clients for all microservices with portal extension dependents
  - [ ] Run `dotnet beam project generate pe-client --service-name <name>` for a specific service and confirm only that service's dependents are updated
  - [ ] Run with a non-existent `--service-name` and verify a `CliException` is thrown
  - [ ] Verify `dotnet build` on a microservice project triggers the `GeneratePortalExtensionClients` target automatically
  - [ ] Confirm the existing federation data loading (`ReloadFederationsData`) still works correctly after the refactor
  - [ ] Run `dotnet test tests/` to check for regressions

### DIFF
--- a/cli/cli/Commands/Project/GeneratePortalExtensionClientsCommand.cs
+++ b/cli/cli/Commands/Project/GeneratePortalExtensionClientsCommand.cs
@@ -1,4 +1,5 @@
 using Beamable.Server;
+using cli.Dotnet;
 using cli.Services;
 using cli.Services.Web;
 using Microsoft.OpenApi.Models;
@@ -9,7 +10,9 @@ namespace cli.Commands.Project;
 
 public class GeneratePortalExtensionClientsCommandArgs : CommandArgs
 {
-	public string ServiceName;
+	public List<string> services = new List<string>();
+	public List<string> withServiceTags = new List<string>();
+	public List<string> withoutServiceTags = new List<string>();
 }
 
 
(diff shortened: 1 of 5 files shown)


cli/cli/Commands/Project/GeneratePortalExtensionClientsCommand.cs
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	public override async Task Handle(GeneratePortalExtensionClientsCommandArgs args)
+	{
+		var allServices = args.BeamoLocalSystem.BeamoManifest.ServiceDefinitions;
+		var allExtensions = allServices.Where((s) => s.Protocol == BeamoProtocolType.PortalExtension).ToList();
+
+		var micros = new List<BeamoServiceDefinition>();
+
+		if (!string.IsNullOrEmpty(args.ServiceName))
```
gabrielbeamable: Use this:
`ProjectCommand.AddIdsOption(this, (args, i) => args.services = i);
		ProjectCommand.AddServiceTagsOption(this,
			bindWithTags: (args, i) => args.withServiceTags = i,
			bindWithoutTags: (args, i) => args.withoutServiceTags = i);`

and also this:
`ProjectCommand.FinalizeServicesArg(args,
			withTags: args.withServiceTags,
			withoutTags: args.withoutServiceTags,
			includeStorage: false,
			ref args.services);`

to not duplicate code

microservice/microservice/Targets/Beamable.Microservice.Runtime.targets
```diff
@@ -116,6 +116,17 @@
               Condition="'@(_BeamServiceRef)' != '' AND !Exists('%(_BeamServiceRef.OapiPath)')" />
     </Target>
 
+    <!--
+        Generate client .ts files for portal extensions that are dependant of this service
+    -->
+    <Target Name="GeneratePortalExtensionClients"
```
gabrielbeamable, line 122: try doing a dotnet deploy and see if things are going to work fine

PR #4648, opened 2026-06-09
Fix Account window first-run build config setup #4640

### LINKED ISSUE #4640: Account window throws ArgumentNullException when build Pid is unset
Found during a QA pass on the Unity SDK **5.1.0-PREVIEW.RC9** release candidate.

## Symptom

On a freshly-installed project, the editor Account window logs `ArgumentNullException: Value cannot be null. Parameter name: key` while it is open. The exception repeats a few times (typically twice or thrice) across consecutive IMGUI repaints before it stops, rather than logging just once.

Downstream noise from the same root cause: `GUI Error: You are pushing more GUIClips than you are popping` and `Rethrow as ImmediateModeException`. The uncaught exception unwinds out of IMGUI mid-draw, leaving the GUI clip stack unbalanced. These are symptoms, not separate bugs.

## Root cause

`Editor/UI/AccountWindow/AccountWindow_SignedIn.cs:73`

```csharp
if (cli.pidToRealm.TryGetValue(config.Pid, out var realm))
```

`config.Pid` (the build connection's Runtime Pid) is `null` on a freshly-installed project, where the build connection strings have not been set yet. `Dictionary.TryGetValue` throws `ArgumentNullException` on a null key. The call originates in `AccountWindow.Draw_SignedIn()` from `AccountWindow.OnGUI()` (`AccountWindow.cs:101`), so it fires on the repaints that occur while the window is settling.

## Repro

1. Create a fresh Unity project and install the SDK.
2. Log in for the first time against an existing CID (one whose build connection Runtime Pid has not yet been populated).
3. Observe the Account window logging the exception a few times.

A project that has already been used (build connection Pid already set) does not reproduce, since `config.Pid` is non-null. Confirmed not OS-specific (seen on macOS; a re-used project on Windows did not reproduce).

## Suggested fix

Guard the lookup against a null/empty Pid:

```csharp
if (!string.IsNullOrEmpty(config.Pid) && cli.pidToRealm.TryGetValue(config.Pid, out var realm))
```

## Environment

- Unity 6000.3.17f1 (Unity 6.3 LTS)
- Beamable Unity SDK 5.1.0-PREVIEW.RC9 (`com.beamable@f1bd005171cb`)
- macOS

mhijaziB:
Fixes #4640 

https://github.com/beamable/BeamableProduct/issues/4640

# Brief Description

## Summary
- Guard Account window runtime realm lookup when `config-defaults.txt` has no PID.
- Auto-create first-run build config with the production/root PID for single-game accounts.
- For multi-game accounts, wait for game/realm selection before writing build config.
- Import the generated `config-defaults.txt` synchronously so Unity sees the new Resources asset.

## Testing
- Verified fresh single-game login creates `config-defaults.txt` with the production/root PID.
- Verified multi-game login writes build config after selecting a game/realm.
- Verified missing `config-defaults.txt` no longer throws `ArgumentNullException` in the Account window.

# Checklist

# Notes

### DIFF
--- a/client/Packages/com.beamable/Editor/BeamEditor.cs
+++ b/client/Packages/com.beamable/Editor/BeamEditor.cs
@@ -685,13 +685,14 @@ namespace Beamable
 
 			if (writeConfig)
 			{
+				var wasNewConfigFile = !File.Exists(path);
 				string directoryName = Path.GetDirectoryName(path);
 				if (!string.IsNullOrWhiteSpace(directoryName))
 				{
 					Directory.CreateDirectory(directoryName);
 				}
 
-				if (File.Exists(path))
+				if (!wasNewConfigFile)
 				{
 					var fileInfo = new FileInfo(path);
 					fileInfo.IsReadOnly = false;
@@ -708,25 +709,28 @@ namespace Beamable
 				}
 				
 				File.WriteAllText(path, asJson);
-				
-				// Import synchronously so Unity's AssetDatabase can see the newly-created
-				// Resources asset before callers continue. This matters for first-run setup,
-				// where config-defaults.txt may not have existed a moment earlier.
-				AssetDatabase.ImportAsset(path, ImportAssetOptions.ForceSynchronousImport | ImportAssetOptions.ForceUpdate);
-				
-				// ImportAsset can still fail to resolve the asset immediately in some editor
-				// timing cases, so fall back to a synchronous refresh before reloading config.
-				if (AssetDatabase.LoadAssetAtPath<TextAsset>(path) == null)
+				if (wasNewConfigFile)
 				{
-					AssetDatabase.Refresh(ImportAssetOptions.ForceSynchronousImport);
+					// Import synchronously so Unity's AssetDatabase can see the newly-created
+					// Resources asset before callers continue. This matters for first-run setup,
+					// where config-defaults.txt may not have existed a moment earlier.
+					AssetDatabase.ImportAsset(path, ImportAssetOptions.ForceSynchronousImport | ImportAssetOptions.ForceUpdate);
+
+					// ImportAsset can still fail to resolve the asset immediately in some editor
+					// timing cases, so fall back to a synchronous refresh before reloading config.
+					if (AssetDatabase.LoadAssetAtPath<TextAsset>(path) == null)
+					{
+						AssetDatabase.Refresh(ImportAssetOptions.ForceSynchronousImport);
+					}
+
+					// Keep the Project window in sync for editor-driven writes, but avoid repainting
+					// editor UI while Unity is in the build pipeline.
+					if (!BuildPipeline.isBuildingPlayer)
+					{
+						EditorApplication.RepaintProjectWindow();
+					}
 				}
 				
-				// Keep the Project window in sync for editor-driven writes, but avoid repainting
-				// editor UI while Unity is in the build pipeline.
-				if (!BuildPipeline.isBuildingPlayer)
-				{
-					EditorApplication.RepaintProjectWindow();
-				}
 				ServiceScope.GetService<ConfigDatabaseProvider>().Reload();
 			}
 		}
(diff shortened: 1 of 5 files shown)


client/Packages/com.beamable/Editor/Toolbar/MenuItems/BeamableRealmMenuItem.cs
```diff
@@ -86,7 +86,8 @@ public override void ContextualizeMenu(BeamEditorContext editor, GenericMenu men
 				});
 				if(!sameEditorAndBuildPids)
 				{
-					menu.AddItem(new GUIContent(rootDisplay.text + "/Save to config-defaults"), false, editor.CommitConfigDefaults);
+					menu.AddItem(new GUIContent(rootDisplay.text + "/Save to config-defaults"), false, 
+					             () => editor.CommitConfigDefaults());
```
allister-beamable, line 90: This seems unnecessary; is there some benefit to constructing a new lambda as opposed to passing the function directly?

client/Packages/com.beamable/Editor/UI/AccountWindow/AccountWindow_SignedIn.cs
```diff
@@ -69,14 +70,14 @@ void RenderLabel(string label, string value, string placeholder = "")
 						RenderLabel("Runtime Host", config.HostUrl, "No value set");
 						RenderLabel("Runtime Cid", config.Cid, "No value set");
 
-						var realmDisplay = config.Pid;
-						if (cli.pidToRealm.TryGetValue(config.Pid, out var realm))
+						string realmDisplay = config.Pid;
```
allister-beamable: When in Rome, do as the Romans do. We shouldn't be changing `var` to `string` unless the official code style says so.

client/Packages/com.beamable/Editor/BeamEditor.cs
```diff
@@ -684,8 +706,27 @@ public void CommitConfigDefaults()
...
+				{
+					AssetDatabase.Refresh(ImportAssetOptions.ForceSynchronousImport);
+				}
+				
+				// Keep the Project window in sync for editor-driven writes, but avoid repainting
+				// editor UI while Unity is in the build pipeline.
+				if (!BuildPipeline.isBuildingPlayer)
+				{
+					EditorApplication.RepaintProjectWindow();
+				}
```
allister-beamable: How much time and memory does this cost? Are we fixing a null ref by adding a bunch of bloat? Maybe this is fine, and maybe it needed to happen anyhow, but I want to make sure we are very intentional and judicious about _anything_ that might slow down initialization.

allister-beamable (changes requested):
Some feedback, of varying degrees:
- why the lambda on line 89?
- why change `var` to `string`?
- why add AssetDatabase stuff?

Nothing too major, but code hygiene is important, and thinking about performance is important.

client/Packages/com.beamable/Editor/Toolbar/MenuItems/BeamableRealmMenuItem.cs
mhijaziB: The `CommitConfigDefaults` now has the `pidOverride` optional parameter. The menu callback expects a parameter-less delegate and the method group no longer binds cleanly, so the lambba preserves the existing behavior.

client/Packages/com.beamable/Editor/UI/AccountWindow/AccountWindow_SignedIn.cs
mhijaziB: Fair. I changed this one intentionally because the null guard made the type matter more, i'll switch the `string` back to `var`

client/Packages/com.beamable/Editor/BeamEditor.cs
mhijaziB: That is a fair point, these were added since I observed that creating the config-defaults wasn't as consistent. I stress test this.
After adding these, creation is very consistent. 
I took into consideration that this method is used in other places and added some guard.
But i can narrow it and for the import/refresh ONLY when the file doesn't exist.

allister-beamable (approved):
With changes, looks good. I would like to hear from others with more SDK experience like @cdhanna and @Leinnan

Leinnan (approved):

PR #4651, opened 2026-06-10
Fix/4645

### LINKED ISSUE #4645: BeamContext init retry loop overflows its errors array and throws IndexOutOfRangeException
## Summary

The retry harness in `BeamContext.Init` (the local `Try()` coroutine) allocates its `errors` buffer sized to `ContextRetryDelays.Length`, but indexes it with the unbounded `attempt` counter. When `EnableInfiniteContextRetries` is `true` (the default) and `InitProcedure()` keeps failing, `attempt` climbs past the array length and the error handler throws `IndexOutOfRangeException` from inside its own `.Error(...)` callback.

This masks the real underlying init failure: instead of surfacing the actual error that is causing init to retry, the context dies with a confusing `IndexOutOfRangeException` originating in Beamable SDK code.

## Where

`client/Packages/com.beamable/Runtime/BeamContext.cs` (line numbers from `main`):

```csharp
var attemptDurations = coreConfig.ContextRetryDelays;
var errors = new Exception[attemptDurations.Length];   // line 471: sized to delays length
while (!success)
{
    var attemptIndex = attempt;
    yield return InitProcedure()
        .Error(err =>
        {
            errors[attemptIndex] = err;                // line 478: attemptIndex == attempt, unbounded
            ...
        })
        ...
    if (attempt >= attemptDurations.Length - 1 && !coreConfig.EnableInfiniteContextRetries)
    {
        _initPromise.CompleteError(new BeamContextInitException(this, errors));
        yield break;
    }
    var duration =
        attemptDurations[attempt >= attemptDurations.Length ? attemptDurations.Length - 1 : attempt];  // guarded
    ...
    attempt++;
}
```

Note that the **delay lookup is already guarded** against running past the array (`attempt >= attemptDurations.Length ? attemptDurations.Length - 1 : attempt`), but the **`errors[attemptIndex]` write is not**.

## Conditions

- `CoreConfiguration.EnableInfiniteContextRetries` defaults to `true`.
- `CoreConfiguration.ContextRetryDelays` defaults to 13 entries, so `errors` has length 13.
- With infinite retries on, when init fails repeatedly the loop never hits the `yield break` at line 497, and `attempt` keeps incrementing. On the **14th attempt (`attemptIndex == 13`)**, `errors[13]` throws `IndexOutOfRangeException`.

## Impact

Any persistent init failure (for example, a refresh-token exchange that keeps returning 401) is turned into an unrelated `IndexOutOfRangeException` thrown from within `BeamContext`'s retry handler after the 14th attempt. This is misleading during diagnosis - the visible exception points at SDK internals rather than the real failure - and it means the configured "infinite retry" behavior is not actually safe past `ContextRetryDelays.Length` attempts.

## Observed stack trace (from a support case, com.beamable@5.0.1)

```
IndexOutOfRangeException: Index was outside the bounds of the array.
(wrapper stelemref) System.Object.virt_stelemref_class_small_idepth(intptr,object)
Beamable.BeamContext+<>c__DisplayClass108_2.<Init>b__1 (System.Exception err) (at ./Library/PackageCache/com.beamable@5.0.1/Runtime/BeamContext.cs:476)
Beamable.Common.Promise`1[T].CompleteError (System.Exception ex) (at ./Library/PackageCache/com.beamable@5.0.1/Common/Runtime/Promise.cs:216)
...
```

## Suggested fix

Either guard the `errors` write the same way the delay lookup is guarded, or grow/clamp the buffer so the unbounded `attempt` index cannot overflow it - e.g. index with `Math.Min(attemptIndex, errors.Length - 1)`, or store errors in a growable list. The underlying init error should remain visible rather than being replaced by an `IndexOutOfRangeException`.

## Affected versions

Present in `com.beamable@5.0.1` and still present on `main` and `unity-sdk-5.1.0-PREVIEW.RC9` as of filing.

## How it was found

Discovered while diagnosing a customer support case where `BeamContext` init kept failing with a 401 during the refresh-token exchange; the repeated failures eventually surfaced as this `IndexOutOfRangeException` instead of the real error.

mhijaziB:
Fixes #4645 

## Summary
- Fixes BeamContext infinite retry handling so retry attempts beyond the configured retry-delay array do not overflow the tracked error buffer.
- Adds PlayMode coverage for persistent init failure with infinite retries enabled.

## Validation
- Unity Test Runner, PlayMode:
  - `Tests.Runtime.Beamable.BeamContextInitTests.InfiniteRetriesDoNotOverflowErrorBuffer`
  - Passed: 1, Failed: 0

## Notes
- The test intentionally mocks `/basic/accounts/me` to return `401` repeatedly.
- The expected behavior is that retries continue, `OnReady` remains pending, and no `IndexOutOfRangeException` is logged.

# Checklist

# Notes


allister-beamable:
I think GitHub specifically looks for "Fixes", "Resolves", or "Closes" as opposed to "Fixing" 😞
The lengths we must go to, to make the robots happy! 🥴

allister-beamable:
Or maybe I'm wrong and it does honor the "-ing" form. Did you link this PR to the issue by hand, or did GH do it automatically?

mhijaziB:
@allister-beamable Yeah, i linked the ticket and this PR to each other. I changed it to fixes and it seems to have pulled it up correctly 
image

client/Packages/com.beamable/Runtime/BeamContext.cs
allister-beamable, line 509: Stepping back a meter or two, why are we using an array for this at all? Why not a list? Doesn't seem like it is in such a performance-critical spot that the overhead of `List<Exception>` and `.Add()` would really matter, and the ergonomics of `List` in this context are much better in my opinion.

allister-beamable: From an even further remove: what is the value of keeping a sequence of errors and a sequence of durations anyhow? Does it present valuable information to the game developer? Does it present valuable information to Beamable staff if the developer shares the error log? If that value is low or nonexistent, I say we ditch the idea of gathering these sequences entirely.

At a certain point, retrying becomes pretty useless except as a "poll until transient conditions vanish" mechanism.

allister-beamable (changes requested):
Some philosophical questions. I do not want to get _too_ bogged down in them, but I think it is worth pausing for a moment to think about system quality as a whole. If there are strong arguments for keeping things as they are, I am willing to be overruled. Let's discuss a little and see if we arrive at consensus (and if not, compromise is good enough).

client/Packages/com.beamable/Runtime/BeamContext.cs
mhijaziB: Both your points are valid but I believe they might go beyond the scope of this specific bug. I kept the existing fixed-size behavior because `BeamContextInitException.Exceptions` is currently documented and tested as matching `ContextRetryDelays.Length`

As for the "infinite retry`, I believe it would be best to discuss this separately with @cdhanna as he has the most SDK knowledge.

allister-beamable (approved):
No need to hold up the PR on the sort of philosophy I raised earlier. This'll be better than what came before it, for sure.

PR #4661, opened 2026-06-12
Add RC conveniences to release-unity.yml

allister-beamable:
Folds the useful parts of the (now retired, #4660) combined RC workflow into the real **Release Unity SDK** workflow — keeping CLI and Unity decoupled, and keeping the npm trusted publisher intact (this is the trusted filename, so public-npm publishing keeps working).

## Changes

**1. `nugetVersion` input (the deep-file fix + reuse-prior-CLI)**
Optional. When set, jq-writes `nugetPackageVersion` in `versions-default.json` at build time — no more hand-editing that deeply-nested file. **Blank keeps the SDK's current dependency**, i.e. reuse the previous CLI artifact when the CLI has not changed (per Chris). Provide a value only when bundling a freshly cut CLI.

**2. Auto RC number**
For `releaseType: rc`, `rcNumber: 0` (the default) derives the next RC from existing `unity-sdk-<maj.min.patch>-PREVIEW.RC*` tags (read via `git ls-remote`, so a shallow clone is fine). An explicit `rcNumber > 0` always wins. Cut RC2..N without tracking the number.

**3. Fail-fast changelog guard**
For `rc`/`production`, the changelog's top `## [x.y.z]` entry must match `major.minor.patch`, checked right after checkout before any build. nightly/exp are unaffected.

## Notes
- Non-rc modes are otherwise unchanged.
- The orphaned `com.beamable@5.1.0-PREVIEW.RC11` on Nexus (from the retired workflow) means auto-RC would compute `11` and collide. Cut the next Unity RC with **`rcNumber: 12`** explicitly until that line moves on.

### DIFF
--- a/.github/workflows/release-unity.yml
+++ b/.github/workflows/release-unity.yml
@@ -24,9 +24,14 @@ on:
         default: 0
       rcNumber:
         type: number
-        description: Rc version number, if required, 0.0.0-PREVIEW.RCX
+        description: "Rc version number (0 = auto: next RC after the latest tag)"
         required: false
         default: 0
+      nugetVersion:
+        type: string
+        description: "CLI nuget version to bundle. Blank = keep the SDK's current dependency."
+        required: false
+        default: ""
       releaseType:
         type: choice
         description: What type of release is this?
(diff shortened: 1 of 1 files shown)


Leinnan:
@allister-beamable Do you want to update this and then merge it? I think I am fine with the change since if I got this right when I input the RC number myself this change does not affect me?

PR #4669, opened 2026-06-15
Remove busted unified Unity+CLI RC publish workflow

allister-beamable:
## What

Removes `.github/workflows/release-unity-rc.yml` (the unified "RC Publish (Unity + CLI) for testing" workflow added in #4655).

## Why

The **"Release com.beamable UPM Package to Public"** step publishes to `https://registry.npmjs.org` with no auth token (`NODE_AUTH_TOKEN`/`NPM_AUTH` are absent, unlike the adjacent Nexus publish steps), so the workflow fails on an npmjs.org permissions error. Removing it so nobody mistakenly tries to use the broken unified workflow.

## Fallback

No gap — RCs continue to be cut via the existing workflows:
- **Unity SDK:** `release-unity.yml` with `releaseType=rc`
- **CLI:** `release.yml`

## Note

Not urgent to land. The 5.1.0 RC is being cut via the two-step path above; this cleanup should merge once that RC is published.


allister-beamable:
I had wanted to have a single button-push for RCs, but that proved to be more trouble than it was worth. The `npmjs.org` publish is keyed to release-unity.yml specifically, so it would always fail from a combined or aggregated job.

PR #4693, opened 2026-06-17
Credentials dropdown in sign in window

### LINKED ISSUE #4776: Credentials dropdown in sign in window
## Reason for the Feature

It makes it easier to login to the places game maker was logged in before. 

Adds dropdown menu with saved credentials to choose from.

Leinnan:
Adds dropdown menu with saved credentials to choose from. Resolves #4776 

image

### DIFF
--- a/client/Packages/com.beamable/Editor/Toolbar/BeamableToolbarExtender.cs
+++ b/client/Packages/com.beamable/Editor/Toolbar/BeamableToolbarExtender.cs
@@ -82,13 +82,16 @@ namespace Beamable.Editor.ToolbarExtender
 			try
 			{
 				var badgeColor = new Color(0,0,0,.3f);
-				if (_editorAPI?.BeamCli.CurrentRealm?.IsProduction ?? false)
+				if (!_editorAPI?.BeamCli?.IsLoggedOut ?? true)
 				{
-					badgeColor = new Color(1, 0, 0, .5f);
-				}
-				else if (_editorAPI?.BeamCli.CurrentRealm?.IsStaging ?? false)
-				{
-					badgeColor = new Color(1, .5f, 0, .5f);
+					if (_editorAPI?.BeamCli.CurrentRealm?.IsProduction ?? false)
+					{
+						badgeColor = new Color(1, 0, 0, .5f);
+					}
+					else if (_editorAPI?.BeamCli.CurrentRealm?.IsStaging ?? false)
+					{
+						badgeColor = new Color(1, .5f, 0, .5f);
+					}
 				}
 
 				var realmDisplay = _editorAPI?.BeamCli.CurrentRealm?.DisplayName ?? "<no realm>";
@@ -107,11 +110,11 @@ namespace Beamable.Editor.ToolbarExtender
 					versionDisplay = "nightly";
 				}
 
-				var loggedOutText = _editorAPI?.BeamCli?.IsLoggedOut ?? false
-					? "[Logged out] "
-					: "";
+				var titleBaseText = _editorAPI?.BeamCli?.IsLoggedOut ?? false
+					? "[Logged out]"
+					: realmDisplay;
 				
-				var titleContent = new GUIContent(loggedOutText + realmDisplay + " (" + versionDisplay + ")");
+				var titleContent = new GUIContent(titleBaseText + " (" + versionDisplay + ")");
 
 				GUI.enabled = _editorAPI != null;
 				var didClick = GUILayout.Button(titleContent, new GUIStyle(EditorStyles.toolbarButton)
--- a/client/Packages/com.beamable/Editor/UI/AccountWindow/AccountWindow_SignIn.cs
+++ b/client/Packages/com.beamable/Editor/UI/AccountWindow/AccountWindow_SignIn.cs
@@ -43,6 +43,19 @@ namespace Beamable.Editor.Accounts
 			envMap[Env.Custom],
 		};
 
+		[Serializable]
+		public class SavedCredentials
+		{
+			public string CidOrAlias;
+			public string Email;
+		}
+
+		[Serializable]
+		class SavedCredentialsArray
+		{
+			public SavedCredentials[] items;
+		}
+
 		public string GetHostString()
 		{
 			switch (env)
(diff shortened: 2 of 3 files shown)


client/Packages/com.beamable/Editor/Toolbar/BeamableToolbarExtender.cs
```diff
@@ -82,13 +82,16 @@ static void OnGUI(IMGUIContainer container)
 			try
 			{
 				var badgeColor = new Color(0,0,0,.3f);
-				if (_editorAPI?.BeamCli.CurrentRealm?.IsProduction ?? false)
+				if (!_editorAPI?.BeamCli?.IsLoggedOut ?? true)
```
allister-beamable, line 85: Ugh, I lowkey hate this syntax. I suppose it is necessary because the value of `_editorAPI?.BeamCli?.IsLoggedOut` ends up being `null` if any of the question-marked ones is null?

No need to change it, just grumping about C# syntax.

client/Packages/com.beamable/Editor/UI/AccountWindow/AccountWindow_SignIn.cs
```diff
@@ -55,6 +68,29 @@ public string GetHostString()
...
+						EditorPrefs.GetString("Beamable.SavedCredentials", "{}")).items.ToList();
+				}
+				catch
+				{
+					return new List<SavedCredentials>();
+				}
+			}
+			set
+			{
+				var vv = JsonUtility.ToJson(new SavedCredentialsArray{items = value
```
allister-beamable, line 88: Not thrilled about `vv` as a variable name; I suppose its only use is right next to it, so it is not actually all that unclear, and `value` is a keyword. Were it my code, I would likely name it `credsValue` or something. Verbosity in hopes of clarity.

No real need to change, just a statement of code style preference: I would rather see longer, more descriptive variable names.

allister-beamable (approved):
Some style/syntax comments, but nothing really demanding changes. Otherwise looks good!

Any concerns about what happens when there is an egregiously large set of saved credentials? I suppose if the main symptom is that the rendered list threatens to overflow the available space, the people adding credentials will notice and perhaps avoid adding more. 😸

I would like to hear from @cdhanna and @DiasAtBeamable on this. Feels like we can take our time with it since it is not harming anything not to have it.

client/Packages/com.beamable/Editor/UI/AccountWindow/AccountWindow_SignIn.cs
```diff
@@ -55,6 +68,29 @@ public string GetHostString()
...
 		}
 
+		List<SavedCredentials> EditorSavedCredentials
+		{
+			get
+			{
+				try
+				{
+					return JsonUtility.FromJson<SavedCredentialsArray>(
+						EditorPrefs.GetString("Beamable.SavedCredentials", "{}")).items.ToList();
```
cdhanna: Unless we have all the tooling for clearing out the editor prefs- I am sad about mixing our data stores. 

Is the goal to remember credentials from between all your projects, or just this one Unity project? If it is just the one unity project, I would recommend we keep our own json file in the .beamable folder. My thinking there is; from the users' perspective, there should be one place where persistent stuff is stored.

client/Packages/com.beamable/Editor/UI/BeamGUI/BeamGUI_PlaceholdTextField.cs
```diff
@@ -45,17 +56,102 @@ public static string PlaceholderTextField(Rect rect, string text, string placeho
...
+            for (int i = 0; i < options.Length; i++)
+            {
+                menuContent[i] = new GUIContent(options[i]);
+                if (options[i] == text)
+                {
+	                selected = i;
+                }
+            }
+
+            EditorUtility.DisplayCustomMenu(
```
cdhanna, line 136: if possible, I would vote we re-use the existing beam editor dropdown window stuff we use in other parts of the editor; rather than use the right-click visual style. I think we could control the UI a little better to even group accounts by their alias and such.

Leinnan: I went this way because the IMGUI with dropdown had issues with depth sorting, but I am open for other approach, more consistent with the rest of the UI

client/Packages/com.beamable/Editor/UI/AccountWindow/AccountWindow_SignIn.cs
Leinnan: Oh, I forgot that the `EditorPrefs` is shared between projects. I am not so confident in using `.beamable` for that though, what do You thinkg about using `EditorUserSettings` `GetConfigValue` and `SetConfigValue`? Then it would be stored in the `%PROJECT_ROOT%/UserSettings/EditorUserSettings.asset` which is fine for me, since I decided not to store the password there.

client/Packages/com.beamable/Editor/Toolbar/BeamableToolbarExtender.cs
Leinnan: 🫂

client/Packages/com.beamable/Editor/UI/AccountWindow/AccountWindow_SignIn.cs
Leinnan: FIxed

client/Packages/com.beamable/Editor/UI/AccountWindow/AccountWindow_SignIn.cs
```diff
@@ -55,6 +68,30 @@ public string GetHostString()
 			}
 		}
 
+		List<SavedCredentials> EditorSavedCredentials
+		{
+			get
+			{
+				try
+				{
+					var savedValue = EditorUserSettings.GetConfigValue("Beamable.SavedCredentials");
```
DiasAtBeamable, line 77: Would be nice to have "Beamable.SavedCredentials" as a const string variable.

DiasAtBeamable (approved):
LGTM, I've just added a suggestion.

allister-beamable:
Let's create a GitHub Issue to act as an anchor point for this.

Leinnan:
> Let's create a GitHub Issue to act as an anchor point for this.

Done

allister-beamable:
Moving to Unity 6.3.0 milestone for further examination: this is a complex area just due to the different ways that "logged in" and "realm selected" are expressed in different contexts. Merits re-examination.

PR #4694, opened 2026-06-17
Make Accounts.OnReady resilient to invalid remembered device tokens

### LINKED ISSUE #4650: Make Accounts.OnReady resilient to stale or invalid remembered device tokens
## Describe the Bug

`BeamContext.Accounts.OnReady` can fail completely when one remembered device token in local storage is stale, invalid, revoked, unknown, or scoped to the wrong realm.

`PlayerAccounts.PerformRefresh()` loads remembered device tokens from `AccessTokenStorage.RetrieveDeviceRefreshTokens(cid, pid)`, appends the current token, calls `GetUser()` for each token, and awaits them via `Promise.Sequence`. Because this is fail-fast, a single bad remembered token can prevent the entire Accounts layer from becoming ready, even when the active `BeamContext` has a valid player.

This is especially risky after realm switching or demo/full-game flows where local token state may contain stale tokens from a previous realm.

1. Initialize Beamable and wait for `BeamContext.Default.OnReady`.
2. Inject or retain a bad remembered device token for the active `cid/pid`
3. Await `BeamContext.Default.Accounts.OnReady`.
4. Observe that `Accounts.OnReady` throws due to the bad remembered token.

Example repro shape:

```csharp
await BeamContext.Default.OnReady;

var ctx = BeamContext.Default;
var badDeviceTokensKey = $"device-tokens{ctx.Cid}{ctx.Pid}";
PlayerPrefs.SetString(badDeviceTokensKey, "bad_access_token|bad_refresh_token");
PlayerPrefs.Save();

await ctx.Accounts.OnReady; // throws
```
Related real-world scenario: after switching realms in the Unity Editor, a realm-scoped token from one realm may be presented while the SDK is operating in another realm. The backend correctly rejects that token, but the SDK should not allow one bad remembered token to break `Accounts.OnReady`.

## Expected Behavior
`Accounts.OnReady` should tolerate bad remembered device tokens.
If a remembered token fails validation, the SDK should skip, remove, or quarantine that token and continue initializing accounts from remaining valid tokens. A stale remembered token should not prevent the Accounts layer from becoming ready when the current context token is valid.
Hard failure should be reserved for cases where the active/current player token cannot be refreshed or validated and there is no safe recovery path.

## Actual Behavior
`Accounts.OnReady` fails when any remembered token fails `GetUser()`.
Because `PlayerAccounts.PerformRefresh()` awaits all remembered-token `GetUser()` calls with `Promise.Sequence`, one invalid token causes the whole refresh to fail. This can surface as `InvalidTokenError` or `UnknownTokenError` from `/basic/accounts/me`.

## Metadata
- SDK version: 5.0.1
- Engine version: Unity 6000.3.x
- Operating System: Windows 11

## Additional Context
This is separate from, but can combine badly with, issue `#4645 (IndexOutOfRangeException in the retry path)`. `#4645` masks persistent initialization/auth failures and creates log spam. This ticket is specifically about making `Accounts.OnReady` resilient to stale remembered device tokens.
Known workaround: clear local token state / PlayerPrefs and restart Unity or the game process. For demo/full-game Steam flows, using separate Unity ProductName values can isolate PlayerPrefs between builds. However, the SDK should still defend against stale remembered-token entries.

mhijaziB:
Fixes #4650 

## Summary

`PlayerAccounts.PerformRefresh()` used to load remembered device tokens, append the current token, and validate all of them through one fail-fast `Promise.Sequence`. That meant one stale remembered token could fail `Accounts.OnReady`, even when the active `BeamContext` player token was valid.

This changes the refresh flow so the current token remains strict, while remembered device tokens are treated as best-effort account candidates.

## Changes

- Validate the current token first and fail cleanly if it is missing, incomplete, or rejected.
- Fetch the current user and current stats outside remembered-token recovery logic.
- Refresh remembered tokens separately and in parallel.
- Remove remembered tokens only for terminal token/auth failures or malformed local token entries.
- Retain remembered tokens on non-terminal failures by default.
- Treat remembered stats failures as non-fatal and fall back to empty stats.
- Skip remembered tokens that duplicate the current refresh token.
- Persist remembered-token cleanup with a single `PlayerPrefs.Save()` after removals.

## Rationale

We want to preserve the existing strict behavior for the active session. If the current player token is bad, `Accounts.OnReady` should still fail because we cannot safely initialize the current account.

Remembered device tokens are different. They are local convenience entries that may become stale after realm switching, demo/full-game flows, or local PlayerPrefs reuse. One bad remembered token should not block the active player from reaching `Accounts.OnReady`.

We also kept performance in mind. Remembered-token refreshes are still queued before `Promise.Sequence`, so multiple remembered accounts do not add avoidable serial latency. Malformed remembered tokens are removed before any network call.

## Validation

- Manually verified valid current token + bad remembered token:
  - `BeamContext.OnReady` succeeds.
  - `Accounts.OnReady` succeeds.
  - bad `device-tokens...` entry is removed.
- Verified corrupting the active/current token fails before or during `BeamContext.OnReady`, which confirms that current-token strictness is still a separate path.

# Checklist

# Notes


client/Packages/com.beamable/Runtime/Player/PlayerAccounts.cs
```diff
@@ -2013,6 +2013,98 @@ private IStatsApi GetStatsServiceForAccount(PlayerAccount account)
...
+			try
+			{
+				var user = await _authService.GetUser(token);
+				UserExtensions.StatCollection statValues;
+
+				try
+				{
+					statValues = await GetStatsForUser(user);
+				}
+				catch
```
allister-beamable, line 2171: I am leery of this bare `catch`; will it populate empty stats if the GetStatsForUser call hits an exception because of a transient 504? What about other exceptions?

client/Packages/com.beamable/Runtime/Player/PlayerAccounts.cs
```diff
@@ -2028,33 +2120,81 @@ protected override async Promise PerformRefresh()
...
+				if (result.account != null)
+				{
+					refreshedAccounts.Add(result.account);
+				}
+
+				if (result.tokenToRemove != null)
+				{
+					_storage.RemoveDeviceRefreshToken(_ctx.Cid, _ctx.Pid, result.tokenToRemove);
+					removedRememberedToken = true;
+				}
```
allister-beamable, line 2283: I do not think we should be removing tokens, even if they are malformed, **unless** that removal is a necessary step to clear the way for a clean login.

My reasoning: there should not be malformed tokens in the first place, so their presence is interesting in terms of forensic diagnostics. If we delete the malformed token, we destroy the evidence, and detectives (people debugging) are sad.

allister-beamable (changes requested):
I would like to strongly challenge the removal part of this. I believe that skipping a bad token is lightweight enough that we can afford to do it every single time the bad token is present, and the downside of deleting "bad" tokens is that we lose a way to diagnose the situation.

We also need to think about this from a higher perspective: this change treats the symptom but does not cure the cause. What changes could we put in place, either in the SDK or in backend or analytics or whatever, to catch and diagnose the stuff that is writing malformed tokens in the first place?

client/Packages/com.beamable/Runtime/Player/PlayerAccounts.cs
mhijaziB: Yes, this is only for remembered tokens if there are no stats this shouldn't break `Accounts.OnReady`, and when this happens then a remembered token if it is retrieved for other accounts it would just show empty stats like display name and stuff like that.

client/Packages/com.beamable/Runtime/Player/PlayerAccounts.cs
mhijaziB: That is something that I personally debated with myself about, but I ended up deciding to remove it my reasoning is that:
1- trying to find the reasons for these is extremely hard, as far as i know, our backend for creating these token is hardened enough. In extreme cases like what happened with Marco, he did something that caused this which is hard to trace.
2- I am still thinking about how we to log these and where.
3- I wanted to open this to debate, so by removing them I knew more people would way in on this.

I do agree though, keeping them could prove more beneficial. I will take a little bit more time into trying to find a way into exposing how was it malformed in the 1st place.

allister-beamable: Good call on point 3! Be provocative, and people **will** respond. 😉

mhijaziB:
@allister-beamable I have updated the PR, please rereview

We now:
- Retain malformed, transient, and unknown remembered tokens.
- Remove only tokens definitively rejected by Beamable.
- Log sanitized diagnostics once per token.
- Keep remembered-account requests parallel.
- Preserve strict current-token behavior.
`remembered_token_failure` is sent only when `CoreConfiguration.enableTokenAnalytics` is enabled. It goes through `Beamable’s AnalyticsTracker` to the Beamable Platform endpoint:
```
POST /report/custom_batch/{cid}/{pid}/{gamerTag}
```
We currently send it immediately, not through the normal batch queue. It contains the reason, action, stage, status, and only the last four token characters.

allister-beamable (approved):
I like the changes. @cdhanna can you or a delegate of your choice weigh in before we merge this to main?

PR #4697, opened 2026-06-18
Auto update PE dependencies and validate PE names

gabrielbeamable:
## Summary
- Add a `beam portal extension update-toolkit` command that bumps `@beamable/portal-toolkit` across every portal extension and library in the workspace, resolving the target version from npm, an explicit `--version`, or local verdaccio (`--local`).
- Introduce portal-extension name-conflict validation that runs at create time (fail fast), deploy time (hard failure), and manifest-load time (non-fatal warning), ensuring names stay unique across microservices, storages, and portal extensions.
- Fix the local web/dev setup scripts to work under Cygwin terminals (path conversion for native-Windows `docker`/`npm`, `.npmrc` location, NuGet source paths).
- Remove the per-template (`BeamService`, `BeamStorage`, `CommonLibrary`) projects from the CLI solution.

### Portal Extension: `update-toolkit` command
- New `PortalExtensionUpdateToolkitCommand` (registered in `App.cs`) updates the `@beamable/portal-toolkit` dependency in `dependencies`, `devDependencies`, and `peerDependencies` of every portal extension and portal-extension library `package.json`, de-duplicated by path.
- Version resolution: `--local` reads verdaccio's `local` dist-tag, `--version` is validated against npm and verdaccio, and the default uses npm's `latest` dist-tag. `--registry` overrides the verdaccio URL.
- The pure `RewriteToolkitVersion` helper (file rewrite only, no npm install) is split out so it can be unit-tested without a network or node toolchain; a best-effort `npm install` follows and only warns on failure.
- `VersionService` gains `GetNpmPackument(...)` plus an `NpmPackument` model tries, with an opt-out (`throwOnError: false`) for probing verdaccio as afallback.
- `PortalExtensionAddLibraryCommand.LocateLibrary` was generalized into `LocateAllLibraries` (returning name + directory + package.json path); the old single-library lookup is now a   thin wrapper over it.
                                                                                                                                                                                        ### Portal Extension: name validation
- New `PortalExtensionNameValidator` centralizes conflict detection: duplicate portal-extension names, and collisions with microservice/storage names (all share one `BeamoId`          namespace).
- `NewPortalExtensionCommand` validates the chosen name up front (before prompts and the remote portal config fetch) so creation fails early.                                           - `DeploymentService` fails fast before the build loop on any conflict across
- `BeamoLocalSystem` surfaces local-only conflicts (e.g. after a rename) as non-fatal warnings on manifest load.                                                                        
### Cygwin support for local dev scripts                                                                                                                                                - `setup.sh`: use `cygpath -m -s` for the repo root under Cygwin (no `pwd -W`et source to a native Windows path so `dotnet nuget add source` accepts it.
- `setup-web.sh` / `dev-web.sh`: convert the docker-compose path to Windows form for the native-Windows `docker` binary, and point `npm config` at the `%USERPROFILE%\.npmrc` that pnpm reads under Cygwin.

### Misc
- `cli.sln`: drop the `BeamService`, `BeamStorage`, and `CommonLibrary` template projects from the solution.
- Add design/spec docs under `cli/projects/` and new tests covering the valid, and portal extension command behavior.


cli/cli/Services/PortalExtension/PortalExtensionNameValidator.cs
```diff
@@ -0,0 +1,109 @@
...
+	/// Core conflict detector. Returns one human-readable message per conflict found.
+	/// </summary>
+	/// <param name="localPortalExtensions">Names of portal extensions that live in the local workspace.</param>
+	/// <param name="deployedPortalExtensions">Names of portal extensions already deployed to the realm. A deployed
+	/// extension that shares a local extension's name is the SAME extension (an update), not a conflict.</param>
+	/// <param name="serviceAndStorageNames">All microservice and microstorage names that will exist (local and/or deployed).</param>
+	public static List<string> FindConflicts(
+		IReadOnlyCollection<string> localPortalExtensions,
+		IReadOnlyCollection<string> deployedPortalExtensions,
+		IReadOnlyCollection<string> serviceAndStorageNames)
```
gabrielbeamable, line 26: should check with libs as well

PR #4698, opened 2026-06-19
Content property validation and inspector

Leinnan:
As the old TODO stated: `Instead of validating those string values we should have a dropdown with already valid options`. 

What it does it removes what did bother me for quite some time, that validation was changing the content values. In some cases that was leading to the problems and we should not do it IMHO.


allister-beamable (changes requested):
Generally looks good.

I would like to see two more things:
- changelog entry
- test plan <- how should those of us doing QA on this set things up to fully exercise this change?

allister-beamable (approved):
With changelog, this is good to go!

allister-beamable:
Retroactively tracked by #4729 — filing an issue to bridge the original support report to this fix, which merged without a linked issue.

PR #4699, opened 2026-06-19
Update mount list source and add Extension site type

gabrielbeamable:
## Summary
- Surface mount sites declared by local Portal Extensions (`<BeamExtensionSite selector="top|bottom" />`) so child extensions can target them when creating a new extension or listing mount-site options — previously only remote Portal page/component slots were visible.
- Introduce a third mount-site kind, `type: "extension"`, addressed by the **host extension's name** rather than a Portal route.
- Make the remote `extension-pages.json` fetch best-effort: if the catalog is unavailable (404/no network) the command warns and continues with locally-discovered sites instead of failing.
- Point the remote config fetch at the new Portal (`api` → `console`).

### Portal extension mount sites (CLI)
- **`IRemotePortalConfigService.cs`** — `GetRemotePortalConfig` now composes `FetchRemotePortalConfig` (best-effort remote fetch, warns and returns empty config on failure) with `AppendLocalExtensionMountSites` (reads the already-loaded `BeamoManifest`, builds sites from local portal-extension defs, merges them in). Adds:
  - `BuildLocalExtensionMountSites` — pure builder: site `path` = extension name, selectors from the scan, `navContext` from the extension's nav fields.
  - `ScanExtensionSiteSelectors` / `ParseExtensionSiteSelectors` — scan `.tsx/.jsx/.ts` (excluding `node_modules`/`dist`/`assets`/`build`/`.git`) and regex-extract `top`/`bottom` selectors, emitted as `#top`/`#bottom` with `type: "extension"`.
  - `MergeMountSites` — merges generated sites by `path`, de-duping selectors, to guard the rare path clash that would otherwise collide last-writer-wins downstream.
- **`ListPortalExtensionOptionsCommand.cs`** — new `extensionMountSites` result category (`ExtensionMountSiteOption { extensionName, selectors }`); sites tagged `type: "extension"` route here, page/component handling unchanged.
- **`NewPortalExtensionCommand.cs`** — `BuildMountSiteIndex` now also produces an `extensionSites` dictionary; `ValidateMountArgs` validates `--mount-page`/`--mount-selector` against host extensions and their slots; `RunMountWizard` gains an "Extension" top-level choice (shown only when extension sites exist) to pick a host extension and one of its sites.
- **`PortalCommand.cs`** — `GetPortalBaseUrl` takes an `isNewPortal` flag that derives the base URL via `api` → `console`; used by the config fetch.

### Tests & docs
- **`RemotePortalConfigServiceTests.cs`** (new) — covers parse, scan (ordering + directory exclusions), build (path = name, `extension` type, navContext, skip cases), and merge behaviors.
- **`portal-extension-mount-sites-plan.md`** (new) — design notes for the feature.


gabrielbeamable:
This has some changes in the .sh scripts, but I that is work of another PR, I just needed that commit to keep working, because of branch swapping.

gabrielbeamable:
This PR kinda depends on this other one in the console repo:
https://github.com/beamable/agentic-portal/pull/55

But merging this wouldn't break anything, you just wouldn't be able to create portal extensions targeting components/page mount sites through the command. But you can still change the package.json with whatever you want and it would work.

gabrielbeamable:
This should change to the following:
 - In build time we regex the project and add to the metadata json the information of it's mounting sites
 - we don't need an Extensions type of mounting site, it can be a component

cli/cli/Commands/Portal/PortalCommand.cs
```diff
@@ -26,13 +26,26 @@ public override async Task Handle(PortalCommandArgs args)
 		MachineHelper.OpenBrowser($"{realmUrl}/{qb}");
 	}
 
-	public static string GetPortalBaseUrl(CommandArgs args)
+	public static string GetPortalBaseUrl(CommandArgs args, bool isNewPortal = false)
```
cdhanna: Instead of `isNewPortal`, I'd call it by name, via enum. 

```csharp
public enum PortalType
{
  LegacyPortal, 
  Console
}

```

cli/cli/Commands/Project/NewPortalExtensionCommand.cs
```diff
@@ -59,11 +58,6 @@ public override void Configure()
 				description: "The mount slot on the page. Required for component extensions; omit for page extensions (auto-assigned). Run 'portal extension list-extension-options' to see valid selectors per page"),
 			binder: (args, i) => args.mountSelector = i);
 		
-		AddOption(new Option<string>(
```
cdhanna, line 62: why does this go away?

cdhanna (changes requested):
looks fine, except for the big question on my mind, "Why are you getting rid of the mount nav group label?"

cli/cli/Commands/Project/NewPortalExtensionCommand.cs
gabrielbeamable: I did it to get rid of confusion
The name of the path now is what defines in which group (hub) it will pertain
Do we want to keep the old behavior in the `new portal-extension` command, but change it in the json?
Or am I miss understanding something here? Does the group field have a different purpose?

PR #4701, opened 2026-06-22
Close shipped Unity milestone on production release

allister-beamable:
## What

Adds a step to `release-unity.yml` that closes the matching `Unity X.Y.Z` milestone once a **production** release publishes, and grants the `issues: write` permission the API call requires.

```yaml
- name: Close shipped milestone
  if: ${{ inputs.dryRun == false && inputs.releaseType == 'production' }}
  ...
  gh api -X PATCH ".../milestones/$num" -f state=closed
```

## Why

Closing the release's milestone after it ships is currently manual muscle-memory that lives with the release shepherd. It was missed for `Unity 5.1.0` (shipped 2026-06-16, still open until closed by hand on 2026-06-22). Encoding it removes one hand-step and keeps milestone state honest.

## Behavior / safety

- **Production only**, and skipped on dry runs.
- Matches a milestone by exact title `Unity ${major}.${minor}.${patch}` (the repo's bare `Unity X.Y.Z` convention).
- **Warns, never fails**, when no open milestone matches - it cannot block a release.
- Does not move open issues forward or create the next milestone; those remain deliberate human decisions.

## Status

Draft, opened for discussion. Open questions worth debating:
- Should it live as an inline step here, or a small dependent job?
- Should it also fail (not just warn) if a production release has no milestone, as a hygiene gate?
- Title-matching is brittle if naming drifts - acceptable, or match another way?


cdhanna (approved):
ahaha, as long as we don't have more than 100 open milestones at once, this looks good ;)

allister-beamable:
Retroactive tracking issue for this change: #4730 (filed for paper-trail completeness; this PR is the resolution).

Issue #4599, opened 2026-05-14
Make microservice content resolution resilient to transient 503s


mhijaziB:
## Reason for the Feature
When resolving content from a C# microservice, transient content fetch failures can currently bubble up as unhandled HttpRequestExceptions.

A BPC customer reported intermittent failures while processing content in a foreach loop. Their code iterates through manifest entries and calls ClientContentInfo.Resolve(). Some entries resolve successfully, then one resolve randomly fails with:
```
System.Net.Http.HttpRequestException: Response status code does not indicate success: 503 (Service Unavailable)
```
The current microservice content resolution path eventually calls:
```
DefaultContentResolver.RequestContent(string uri)
in
microservice/microservice/dbmicroservice/IContentResolver.cs
```
That method currently uses HttpClient.GetStringAsync(uri) directly, with no retry/backoff handling for transient HTTP failures.

## Suggested requirements

- Add retry/backoff handling to DefaultContentResolver.RequestContent(string uri).
- Retry only transient HTTP responses, such as:
  - 429 Too Many Requests
  - 500 Internal Server Error
  - 502 Bad Gateway
  - 503 Service Unavailable
  - 504 Gateway Timeout
- Respect Retry-After headers when available.
- Use exponential backoff with jitter to avoid retry bursts.
- Preserve fail-fast behavior for non-transient failures such as 400, 401, 403, and 404.
- Improve final error reporting so failures include:
  - content URI
  - HTTP status code
  - attempt count
  - ideally the content id, when available from the cache/content layer
- Add tests covering:
  - 503 followed by success
  - repeated 503 eventually failing with a useful error
  - non-transient failures not being retried

The ask is not to hide real failures, but to make the content resolution path resilient to short-lived service unavailability and to provide better diagnostics when retries are exhausted.

## Alternatives or Workarounds
Clients can work around this today by adding their own retry logic around each ClientContentInfo.Resolve() call.

**Additional Context***
Observed client error:
```
HandleContentPublish: unhandled exception:
System.Net.Http.HttpRequestException: Response status code does not indicate success: 503 (Service Unavailable).
   at <redacted>.ContentEventSubscriber.CreateOrUpdateAllItemTypes()

```

Client's code:
```
var manifest = await _contentApi.GetManifest("t:items");

foreach (var clientContentInfo in manifest.entries)
{
    var contentInfo = await clientContentInfo.Resolve();

    // process content
}
```

mhijaziB:
[...]

PR #4704, opened 2026-06-23
fix(content): recover poisoned manifest promise after a failed fetch

ali-beamable:
## Problem

In deployed microservices, content/inventory reads start failing spontaneously with `503 (Service Unavailable)` and only recover when the service is restarted. The production stacktrace lands in `FifaCore.Inventory.InventoryManager.ViewToItems` → `Beamable.Common.PromiseExtensions.Recover`.

## Root cause

`CachedContentManifest.WaitForManifest()` (in `microservice/microservice/Content/ContentService.cs`) caches the manifest as a `Promise<ClientManifest>` and only refetched it when that field was **null**:

```csharp
hasNoManifest = _waitForManifest == null;
if (hasNoManifest) RefreshManifest();
```

`RefreshManifest()` builds the promise from `GET /basic/content/manifest` with only `.RecoverFrom404(...)`. A **503** (e.g. a content-publish refresh coinciding with a platform blip) is not a 404, so it propagates and `_waitForManifest` completes in a **failed** state. Because the field is now non-null, it is never discarded — every later `GetContent` re-awaits the same poisoned promise via `WaitForManifest().FlatMap(...)`, so **all** content resolution fails until the process restarts (fresh `ContentService`, which is registered as a singleton) or a later content-publish event happens to succeed. Caller-side retry loops can't help because each retry re-awaits the identical cached failure.

## Fix

Discard a cached promise that has settled in a failed state, so the next caller triggers a fresh fetch:

```csharp
if (_waitForManifest == null || _waitForManifest.IsFailed)
{
    RefreshManifest();
}
```

- `IsFailed` is `done && err != null` — true only for a settled, failed promise. Null, in-flight, and successful promises are untouched, so the happy path and the existing fetched-once / cache-purged-on-publish behavior are unchanged.
- `RefreshManifest()` already replaces `_waitForManifest` under `_manifestLock`; its `!oldPromise.IsCompleted` guard skips success-forwarding for an already-completed (failed) old promise, so there is no double-complete.
- Recovery is on-demand and lock-serialized: concurrent callers share one in-flight refetch.

## Test

Adds `FirstManifestBreaksWith503_RecoversOnNextRequest` to `GetContentTests.cs`: the first manifest fetch returns a non-gateway 503 (which the requester does **not** auto-recover, unlike a gateway auth failure), asserts the first content request fails, then asserts the next request refetches and succeeds.

Verified locally:
- With the fix: all 7 GetContentTests pass (including the existing `FirstManifestBreaksWithAuth`, `CachePurgedOnNewManifest`, `ContentOnlyFetchedOnce`).
- Without the fix (temporarily reverted): the new test fails exactly as the production bug does — the second `GetContent` re-throws the cached 503 — confirming it's a genuine regression guard.

https://claude.ai/code/session_017LNB9NCnEknRSqze3BYgrR

---
_Generated by [Claude Code](https://claude.ai/code/session_017LNB9NCnEknRSqze3BYgrR)_

### DIFF
--- a/microservice/microservice/CHANGELOG.md
+++ b/microservice/microservice/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- C# microservice content resolution now retries transient HTTP failures and no longer caches failed content resolution tasks [#4599](https://github.com/beamable/BeamableProduct/issues/4599)
+
 ## [7.2.0] - 2026-05-23
 
 ### Changed
--- a/microservice/microservice/CHANGELOG.md
+++ b/microservice/microservice/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- C# microservice content manifest fetches now use bounded retries for transient failures, avoiding unbounded refetch loops after failed manifest requests [#4704](https://github.com/beamable/BeamableProduct/pull/4704)
+
 ## [7.2.0] - 2026-05-23
 
 ### Changed
--- a/microservice/microservice/CHANGELOG.md
+++ b/microservice/microservice/CHANGELOG.md
@@ -8,8 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- C# microservice content manifest fetches now use bounded retries for transient failures, avoiding unbounded refetch loops after failed manifest requests [#4704](https://github.com/beamable/BeamableProduct/pull/4704)
-- C# microservice content resolution now retries transient HTTP failures and no longer caches failed content resolution tasks [#4599](https://github.com/beamable/BeamableProduct/issues/4599)
+- Microservice content fetches now retry with bounded exponential backoff and jitter. This applies to both manifest and content entry fetches.
 
 ## [7.2.0] - 2026-05-23
 
--- a/microservice/microservice/Content/ContentService.cs
+++ b/microservice/microservice/Content/ContentService.cs
@@ -145,11 +145,12 @@ namespace Beamable.Server.Content
 	   
 	   public Promise<ClientManifest> WaitForManifest()
 	   {
-		   var hasNoManifest = true;
 		   lock (_manifestLock)
 		   {
-			   hasNoManifest = _waitForManifest == null;
-			   if (hasNoManifest)
+			   // Refetch when there is no manifest yet, or when the cached promise settled in a
+			   // failed state (e.g. a 503 during a refresh). Without the IsFailed check the failed
+			   // promise would be re-awaited forever, breaking all content resolution until restart.
+			   if (_waitForManifest == null || _waitForManifest.IsFailed)
 			   {
 				   RefreshManifest();
 			   }
--- a/microservice/microservice/Content/ContentService.cs
+++ b/microservice/microservice/Content/ContentService.cs
@@ -115,6 +115,9 @@ namespace Beamable.Server.Content
 
    public class CachedContentManifest
    {
+	   private const int MaxManifestAttempts = 3;
+	   private static readonly TimeSpan MaxManifestRetryDelay = TimeSpan.FromSeconds(2);
+
 	   public string _name;
 	   private readonly MicroserviceRequester _requester;
 	   private readonly ConcurrentDictionary<string, ContentReference> _idToContentReference = new ConcurrentDictionary<string, ContentReference>();
@@ -173,17 +176,13 @@ namespace Beamable.Server.Content
 	   }
 
 	   
-      public void RefreshManifest()
+	   public void RefreshManifest()
       {
          lock (_manifestLock)
          {
             var oldPromise = _waitForManifest;
             // reset the manifest promise, so that if a content request comes while a manifest is being retrieved, we wait for the updated content to be served
-            _waitForManifest = _requester.Request<ContentManifest>(Method.GET, $"/basic/content/manifest?id={_name}")
-	            .RecoverFrom404(ex => new ContentManifest
-	            {
-		            id = _name, created = 0, references = new List<ContentReference>()
-	            })
+            _waitForManifest = RequestManifestWithRetries()
 	            .Map(manifest =>
                {
                   _idToContentReference.Clear();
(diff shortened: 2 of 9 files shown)


mhijaziB:
@allister-beamable @ali-beamable 
Updated this PR to address the unbounded retry concern.

We now bound manifest refetch recovery inside the manifest fetch itself:

- 3 total attempts for transient manifest failures
- retries only status `0`, `408`, `429`, `500`, `502`, `503`, `504`
- no retry for `400`, `401`, `403`, or `404`
- keeps `404 -> empty manifest` behavior intact
- uses short exponential backoff with jitter, matching the shape of #4705 without adding new public config or shared retry plumbing

The key behavior change is that a transient `503` can recover within the same `GetContent` call, instead of relying on the next caller to act as the retry mechanism. Caller traffic is not a retry policy. Tiny goblin avoided.

Added tests for:

- `503 -> success` retries and succeeds on the same request
- repeated `503` stops after exactly 3 attempts
- the next request starts a fresh bounded attempt group after exhaustion
- `400` fails fast
- `404` still recovers as an empty manifest and does not retry

Verified:

`dotnet restore microservice/microserviceTests`
`dotnet build --configuration Release --no-restore microservice/microserviceTests`
`dotnet test --configuration Release --no-build microservice/microserviceTests`

Result: `223` passed, `0` failed.

mhijaziB:
Merged the [content-resilience](https://github.com/beamable/BeamableProduct/pull/4705) branch into this one and added stress coverage around the failure mode from the Slack thread.

The important part is that we are now testing the stampede-shaped case directly:

- 100 concurrent `GetContent` calls share one bounded manifest retry group.
- 100 concurrent `GetContent` calls share one resolved-content download/cache task.
- transient `503, 503, success` behavior stays bounded at 3 total attempts, not multiplied per caller.

This gives us coverage for both layers we discussed:

- manifest promise recovery from #4704
- resolved content retry + failed cache task eviction from #4705 / #4599

Verification:

```text
ContentResilienceStressTests
Passed: 2, Failed: 0

Targeted resilience suite
Passed: 27, Failed: 0

Full microservice test suite
Passed: 240, Failed: 0, Skipped: 0

File: microservice/microservice/Content/ContentService.cs [hunk 1]
```diff
@@ -223,10 +223,49 @@ public void RefreshManifest()
...
+					   });
+			   }
+			   catch (RequesterException ex) when (IsTransientManifestFailure(ex.Status) && attempt < MaxManifestAttempts)
+			   {
+				   await Task.Delay(GetManifestRetryDelay(attempt));
+			   }
+		   }
+
+		   // The loop either returns or rethrows the original final/non-transient exception.
+		   throw new InvalidOperationException("Unreachable manifest retry state.");
```
allister-beamable: If this is truly unreachable, why does it exist at all? Does the C# compiler yell if it is absent?

microservice/microservice/Content/ContentService.cs
```diff
@@ -223,10 +223,49 @@ public void RefreshManifest()
...
+	   }
+
+	   private static bool IsTransientManifestFailure(long status)
+	   {
+		   return status is 0 or 408 or 429 or 500 or 502 or 503 or 504;
+	   }
+
+	   private static TimeSpan GetManifestRetryDelay(int attempt)
+	   {
+		   var backoffMilliseconds = 200 * (1 << (attempt - 1));
```
allister-beamable: I think bit shifting is ugly here; would `Math.Pow(2, attempt)` be significantly worse, computationally?

microservice/microservice/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- C# microservice content manifest fetches now use bounded retries for transient failures, avoiding unbounded refetch loops after failed manifest requests [#4704](https://github.com/beamable/BeamableProduct/pull/4704)
+- C# microservice content resolution now retries transient HTTP failures and no longer caches failed content resolution tasks [#4599](https://github.com/beamable/BeamableProduct/issues/4599)
+
```
allister-beamable: Simplify these and make them more lay-friendly. The changelog need not link back to ticket numbers, as the Git history already provides that. The purpose of the changelog is to communicate to Game Developers what they should expect from a change.

Could both of these be a single bullet point that reads something like:
> - Microservice content fetches now retry with bounded exponential backoff and jitter. This applies to both manifest and content entry fetches.

allister-beamable (changes requested):
Minor syntax & hygiene changes.

With this PR covering both manifest and entry fetches now, does that mean #4705 is obsolete?

Thinking through the scenarios, are there any circumstances where this retry logic could/would push the total time for a `[ClientCallable]` invocation past 10 seconds? Whenever a microservice call takes more than 10.0s, the gateway yields a 504 Gateway Timeout response. In the case that inspired this PR, that kind of 504 would be just as bad as the current state of affairs; better to yield a 503 Service Unavailable with specific information at 960ms or whatever.

microservice/microservice/Content/ContentService.cs
mhijaziB: Good call. For the unreachable throw: yes, the compiler needs a terminal return/throw with the current `for` shape, even though the runtime paths either return or rethrow the original exception. I agree it reads awkwardly. I’ll restructure the loop so the retry state is explicit and we do not need an “unreachable” throw.

microservice/microservice/Content/ContentService.cs
mhijaziB: Computationally, there shouldn't be any difference. I will switch to Math.Pow for readability though

microservice/microservice/CHANGELOG.md
mhijaziB: I will consolidate them into one.

mhijaziB:
@allister-beamable

As for #4705, it is obsolete as a separate PR as the changes there have been merged here. I am just keeping it open just in case and will close it once this one is approved and merged.

On the 10s gateway timeout concern: the retry backoff itself should not get close to 10s. Manifest retries add at most about 800ms of delay: 200ms+jitter, then 400ms+jitter. Resolved content retries add the same in the normal case, or up to 4s total delay if two `Retry-After` values are present and each is capped at 2s.

The caveat is request time, not backoff time. If an underlying manifest/content request stalls before failing, then yes, total wall time could still exceed the gateway’s 10s client-call budget. That was technically possible before this retry work too, but retries can make the budget question more visible.

I would not add a silent long timeout here. If we want a hard guarantee that content resolution fails before the gateway does, we should make that an explicit follow-up/change: pass a call budget or cancellation token through manifest/content fetches, or apply a conservative per-attempt timeout for content entry downloads. That is a slightly different behavior change than bounded retry. These could be out of scope of this PR, we can take another look after though.

allister-beamable:
> I would not add a silent long timeout here. If we want a hard guarantee that content resolution fails before the gateway does, we should make that an explicit follow-up/change: pass a call budget or cancellation token through manifest/content fetches, or apply a conservative per-attempt timeout for content entry downloads. That is a slightly different behavior change than bounded retry. These could be out of scope of this PR, we can take another look after though.

Ok, yeah. Reasonable. I think we should harden the guarantee later.

The most common failure modes I expect are:
- content delivery sprinkles a low rate of 503s during a stampede, as flow control / rate limiting
- widespread network or infrastructure failure leads to a steady state of continuous 503s, 504s, or 500s

In the former case, this PR handles it gracefully with the addition of a fraction of a second. In the latter case, the request was doomed anyway, so the client-facing error is _largely_ academic. (but it is the gap implied by that "largely" that I would like to address later in a different issue and PR 😸)

microservice/microservice/Content/ContentService.cs
```diff
@@ -234,24 +234,28 @@ private Promise<ContentManifest> RequestManifestWithRetries()
...
-			   catch (RequesterException ex) when (IsTransientManifestFailure(ex.Status) && attempt < MaxManifestAttempts)
+			   catch (RequesterException ex) when (IsTransientManifestFailure(ex.Status))
 			   {
 				   await Task.Delay(GetManifestRetryDelay(attempt));
 			   }
 		   }
 
-		   // The loop either returns or rethrows the original final/non-transient exception.
-		   throw new InvalidOperationException("Unreachable manifest retry state.");
+		   return await RequestManifestOnceAsync();
```
allister-beamable, line 249: Yes, good. I like the looks of this pattern better, especially as it clarifies what action is happening and how the loop proceeds.

The loop-of-many + one-more pattern feels comfortingly familiar from the land of LISP and tail recursion. <3

allister-beamable (approved):
Good to go. Let's get this into a patch release.

PR #4738, opened 2026-07-06
Unity IAP Purchasing v5 support

### LINKED ISSUE #4646: Support Unity IAP v5
## Background

Unity has [deprecated In-App Purchasing v4](https://docs.unity.com/en-us/iap/upgrade-to-iap-v5): it still works in live builds, but no longer receives bugfixes or improvements. The Beamable Unity SDK's purchaser (`UnityBeamablePurchaser`) is written entirely against the v4 listener/controller API surface, which v5 removes. We should add support for IAP v5 so customers can upgrade.

## Scope

The purchaser class implements `IStoreListener`, so v5 is a **hard compile break**, not a silent behavioral one — the interface it implements no longer exists. Supporting v5 means re-implementing the Unity-IAP boundary of `UnityBeamablePurchaser` around v5's event-driven `StoreController`. The Beamable-side plumbing (`PaymentService`, `CompletedTransaction`, retry/backoff) is provider-agnostic and should be unaffected.

## Compatibility

Existing customers on 4.x must keep working. Follow the pattern already in the file: gate the v5 code path behind a preprocessor symbol (e.g. `UNITY_PURCHASING_5_OR_NEWER`), mirroring the existing `UNITY_PURCHASING_4_6_OR_NEWER` split, so both API generations compile from one codebase. The editor-side detection in `BeamablePurchaserImporter` already treats any non-`2.x` version as the modern flow, so it does not need to change for v5 to be recognized.

Implementation specifics (the v4→v5 API mapping) are in the comments below.

Leinnan:
# Ticket

Fixes #4646

# Brief Description

Support for version 5 of the Unity IAP Purchasing implementation.


client/Packages/com.beamable/Runtime/Modules/Purchasing/UnityBeamablePurchaser.cs
```diff
@@ -240,9 +365,134 @@ public void RestorePurchases()
...
+			{
+				var receipt = JsonUtility.FromJson<UnityPurchaseReceipt>(receiptString);
+				rawReceipt = receipt != null && !string.IsNullOrEmpty(receipt.Payload) ? receipt.Payload : receiptString;
+				InAppPurchaseLogger.Log($"UnityIAP Payload: {rawReceipt}");
+				InAppPurchaseLogger.Log($"UnityIAP Raw Receipt: {receiptString}");
+			}
+			else
+			{
+				rawReceipt = receiptString;
+			}
```
allister-beamable: Why this `else` clause? Isn't the default value for `string` a null? Like, couldn't line 435 initialize to default, obviating this whole block?

allister-beamable (approved):
Generally looks good. A few questions before making this "real":
- do we have docs about how to override this if desired?
- does it still support Unity IAP v4.x? (either way is fine; we should just document it clearly)
- what do game developers need to do to enable and configure everything? does any of that differ from how it was done in UnityIAP v4.x?

client/Packages/com.beamable/Runtime/Modules/Purchasing/UnityBeamablePurchaser.cs
Leinnan: Good call. Fixed.

Leinnan:
> * do we have docs about how to override this if desired?

Not yet, but we should do it once we will have it merged/released.

> * does it still support Unity IAP v4.x? (either way is fine; we should just document it clearly)

Yes.

> * what do game developers need to do to enable and configure everything? does any of that differ from how it was done in UnityIAP v4.x?

Flow for enabling/configuring did not change.

PR #4744, opened 2026-07-08
Fix Snapshot tab hang when snapshots in different realms share a name

### LINKED ISSUE #4743: Content window Snapshot tab hangs at 'Loading snapshots' when snapshots in different realms share a name
### Description

A community user on Unity SDK 5.1.1 reported that the Snapshot tab of the Beam Content window hangs at "Loading snapshots" and never finishes. This happens after publishing content to more than one realm with "On Publish Auto Snapshot Type" set to Both (or Shared), because each publish writes an auto snapshot named `LastPublished-global.json` into that realm's `.beamable/content-snapshots/<pid>/` folder - producing two snapshot files with the same name in different realm folders.

### Steps to Reproduce

1. In Content Settings, set "On Publish Auto Snapshot Type" to Both (or Shared).
2. Publish content to realm A.
3. Switch realms and publish content to realm B. There are now two `LastPublished-global.json` files, one per realm folder.
4. Open the Beam Content window and select the Snapshot tab.

### Expected

The Snapshot tab lists the snapshots from all realm folders, including auto snapshots that share the file name `LastPublished-global.json`.

### Actual

The Snapshot tab shows "Loading snapshots" indefinitely. Renaming either of the identically named snapshot files restores loading.

### Root Cause

In `ContentWindow_SnapshotManager.CacheSnapshots()`:

- The shared and local snapshot lookups are built with `ToDictionary(item => item.Name)`. Snapshot names are not unique across realm folders (auto snapshots are always `LastPublished-<manifestId>`), so `ToDictionary` throws `ArgumentException` on the duplicate key.
- `_gatheringSnapshots` is set to `true` before the fetch and only cleared on the method's last line. The call is fire-and-forget (`_ = CacheSnapshots();`), so the exception is swallowed and the flag stays `true`, which makes `ContentWindow` draw the "Loading snapshots" block forever.

### Fix

Key the snapshot lookups by file path (which is unique, and is already how the adjacent `_allSnapshots` dictionary is keyed), and clear `_gatheringSnapshots` in a `finally` block so a failure during the cache rebuild can no longer wedge the tab.

allister-beamable:
Fixes #4743

### Problem

The Snapshot tab of the Beam Content window hangs at "Loading snapshots" forever when two realm folders contain snapshots with the same file name. This happens routinely once "On Publish Auto Snapshot Type" is enabled and content is published to more than one realm, because every publish writes `LastPublished-<manifestId>.json` into that realm's snapshot folder.

Root cause, in `ContentWindow_SnapshotManager.CacheSnapshots()`:

- The shared and local snapshot lookups were built with `ToDictionary(item => item.Name)`, and snapshot names are not unique across realm folders, so `ToDictionary` threw `ArgumentException`.
- The method is invoked fire-and-forget (`_ = CacheSnapshots();`), so the exception was swallowed, and `_gatheringSnapshots` (cleared only on the method's last line) stayed `true` - which makes the window draw the "Loading snapshots" block on every repaint.

### Fix

- Key the shared/local snapshot lookups by file path, which is unique and matches how the adjacent `_allSnapshots` dictionary was already keyed. The lookups are only consumed via `.Values`, so no other behavior changes.
- Clear `_gatheringSnapshots` in a `finally` block so no failure during the cache rebuild can wedge the tab again.
- Changelog entry under `[Unreleased]`.

### Tests

Added `ContentWindowSnapshotTests` (editor editmode) covering the duplicate-name-across-realms case, path keying, and empty input. Ran locally via Unity batch mode: 3/3 passed.

### DIFF
--- a/client/Packages/com.beamable/CHANGELOG.md
+++ b/client/Packages/com.beamable/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed the Content window Snapshot tab hanging at "Loading snapshots" when snapshots in different realm folders share a name (e.g. `LastPublished-global.json` auto snapshots from publishing to multiple realms) [4743](https://github.com/beamable/BeamableProduct/issues/4743)
+
 ## [5.1.1] - 2026-06-29
 
 ### Added
--- a/client/Packages/com.beamable/CHANGELOG.md
+++ b/client/Packages/com.beamable/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed the Content window Snapshot tab hanging at "Loading snapshots" when snapshots in different realm folders share a name (e.g. `LastPublished-global.json` auto snapshots from publishing to multiple realms) [4743](https://github.com/beamable/BeamableProduct/issues/4743)
+- Fixed the Content window Snapshot tab hanging at "Loading snapshots" when snapshots in different realm folders share a name (e.g. `LastPublished-global.json` auto snapshots from publishing to multiple realms)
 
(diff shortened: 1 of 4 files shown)


mhijaziB:
Simple clean straight forward! Best type of hotfixes.

mhijaziB (approved):

Leinnan (approved):
LGTM

DiasAtBeamable (approved):
LGTM

PR #4745, opened 2026-07-08
Cached OpenAPI docs schema

### LINKED ISSUE #4742: Microservice should cache its OpenAPI document instead of regenerating it on every admin/docs request
## Problem

A C# microservice generates its OpenAPI document by reflecting over all of its registered types (routes plus request/response schemas) to build the schema set. This generation runs in full on **every** `GET admin/docs` request. On CPU-constrained instances, repeated polling of `admin/docs` produces repeated full reflection scans and adds significant, avoidable CPU load.

## Expected

The OpenAPI document cannot change over the lifetime of a running microservice instance -- its routes and types are fixed at build time. After the document is generated once (at or after boot), `admin/docs` should return an identical, cached document without re-running the reflection scan.

## Actual

Each `admin/docs` request re-runs the full reflection walk and rebuilds the document from scratch. Observed in production: one complete reflection pass over the service's schema types (hundreds of types) per request, i.e. one CPU-expensive scan for every hit. A client or probe polling `admin/docs` therefore drives one full scan per poll, which is enough to be a meaningful load on a small instance.

Leinnan:
# Ticket

Resolves #4742

# Brief Description

I've made a simple service that whole caches the ServiceDocument generation output.


allister-beamable (approved):
I think this is worth a try, and would save a fair bit of re-computation of the OpenAPI document whenever Portal hits the docs endpoint.

PR #4752, opened 2026-07-10
Bundles feature

gabrielbeamable:
## Summary
Adds CLI support for Beamo **manifest bundles** — the evolution of the flat per-realm manifest into a manifest-of-manifests that pulls named, content-addressed bundles from a shared catalog. This lands the OpenAPI regen that introduces the bundle/segments/analytics clients (and retypes `ManifestView` onto the new `BeamoBasic*` family), then layers on a `bundles` command group, `manifest.beam.json` reference plumbing, and an internal `admin force-inject` escape hatch.

### OpenAPI regen & build adoption
- Regenerated `Models.gs.cs` plus new generated clients: `BeamBeamobundle`, `BeamBeamoforcedbundle`, `BeamBeamopullrequest`, `BeamSegments`, `BeamCustomersegments`, `BeamPlayersegments`, `BeamAnalytics`
- Repointed `ManifestView` onto the `BeamoBasic*` reference family; added Basic-family overloads to `ConvertToBeamoV2` / checksum helpers and new Basic→canonical `ConvertToBundleReference` converters
- New `OptionalArrayOfArrayOfString` optional type; registered the new bundle clients in `OpenApiSystem.RegisterOpenApis`
- Fully-qualified `System.Net.*` to resolve collisions with new generated `IPAddress`/`AddressFamily` models

### `bundles` command group (`beam bundles <cmd>`)
- `new`, `plan`, `publish`, `list`, `get`, `history`, `tags`, `tag`, `yank`, `pin`, `acl`, `prune-yanked`
- Authored-bundle discovery/validation of `*.beam.bundle.json` files (unique names; each beamoId in at most one bundle)
- Bundle build + diff services (`BundleBuild`, `BundleDiff`, `BundleWorkspace`)
- No-op `@`-token replacer so namespaced names (`<namespace>/<bundle>`) aren't swallowed by System.CommandLine's `@file` response-file expansion

### Manifest references (Phase 1)
- New `manifest.beam.json` (`schemaVersion` + `references` map) with `ConfigService` load/save helpers, wired into `deploy plan`/`release`

### Internal admin
- `beam deploy admin force-inject <name> --realm --checksum [--unset]`, marked internal-use (SuperAdmin-gated server-side)

### OpenAPI generation fix
- Rewrote `SwaggerService` self-reference detection as an explicit path-tracked traversal to stop an infinite loop on recursive schemas

```json
// .beamable/manifest.beam.json
{
  "schemaVersion": 2,
  "references": {
    "<namespace>/<bundle-name>": "sha256:<checksum>"
  }
}

### CODE REVIEW

File: cli/cli/Commands/BundleCommands/BundlePlanCommand.cs [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+
+	public override async Task Handle(BundlePlanCommandArgs args)
+	{
+		var provider = args.DependencyProvider;
+		var bundle = BundleWorkspace.Require(args.ConfigService, args.bundleName);
+		var (ns, name) = BundleWorkspace.SplitBundleName(bundle.name);
+
+		Log.Information($"Building bundle=[{bundle.name}] to compute its diff...");
+		// Don't exclude bundle components here — this command operates on exactly those components.
+		var (plan, _) = await this.InteractivePlan(provider, args, excludeAuthoredBundleComponents: false);
```
gabrielbeamable: we stopped here

cli/cli/Commands/BundleCommands/PinBundleCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+	public PinBundleCommand() : base("pin", "Pin a bundle checksum into the local manifest.beam.json references")
+	{
+	}
+
+	public override void Configure()
+	{
+		AddArgument(new Argument<string>("bundle-name", "The namespaced bundle name, e.g. <namespace>/<bundle-name>"),
+			(args, i) => args.bundleName = i);
+		AddOption(new Option<string>("--tag", () => null, "Resolve the checksum from this tag (defaults to latest)"),
```
gabrielbeamable: the default should be defined here

PR #4754, opened 2026-07-13
Add regression tests for null reference fields in microservice client responses

allister-beamable:
## Summary

Unity SDK 5.1.0 changed the object fallback of `MicroserviceClientHelper.DeserializeResult<T>` from `JsonUtility.FromJson<T>(json)` to SmallerJSON `Json.Deserialize<T>(json)` (74dcc190e). The new mapper preserves explicit JSON nulls as C# `null` and leaves absent members at CLR defaults, where JsonUtility produced non-null defaults. Client code written against 5.0.x relies on those non-null guarantees for response DTO reference fields and now throws NullReferenceException.

This PR adds tests only - it documents the 5.0.x contract so a fix can be developed against it. On this branch the four object-path tests fail; all six pass when the same file is run against `unity-sdk-5.0.1`.

## The 5.0.x contract under test

The default microservice response serializer (Newtonsoft, default `NullValueHandling`) emits unassigned reference fields as explicit JSON nulls, identically on 4.3.8 and current - covered by the new `microserviceTests` case. So the dominant wire shape is explicit-null members:

| Member on the wire | JsonUtility (5.0.x) | SmallerJSON (5.1.0+) |
|---|---|---|
| `"name": null` (string) | `""` | `null` |
| `"roles": null` (array) | `[]` (empty) | `null` |
| `"tags": null` (List) | empty list | `null` |
| `"nested": null` / `"itemRef": null` (object / ContentRef) | default instance | `null` |
| absent `List`, nested object, ContentRef member | default instance | `null` |
| absent `string`, `string[]` member | `null` (unchanged) | `null` |

Absent members occur when a field is a valueless `Optional` or a service hand-serializes its response with null-omitting settings; strings and plain arrays never had a non-null guarantee for absence, so the absent-member test covers only the field types that did.

## Test results

| Test | unity-sdk-5.0.1 | this branch |
|---|---|---|
| Object response, explicit-null reference fields | pass | fail |
| Nested object with explicit-null string member | pass | fail |
| Object response, absent list/nested/ref members | pass | fail |
| List of objects with explicit-null members | pass | pass (list path unchanged) |
| Fully populated object round trip | pass | pass |
| End-to-end request through MicroserviceClient | pass | pass at 5.0.1, fail here |
| Server emits explicit nulls (microserviceTests) | pass at cli-4.3.8 | pass |

JsonUtility reference behavior was verified with a standalone probe on Unity 2021.3.45f2 and 6000.3.12f1 with identical results, so the documented contract is not editor-version specific.

### DIFF
--- a/client/Packages/com.beamable/Tests/Runtime/Server/Runtime/MicroserviceClient/DeserializeResultNullDefaultsTests.cs
+++ b/client/Packages/com.beamable/Tests/Runtime/Server/Runtime/MicroserviceClient/DeserializeResultNullDefaultsTests.cs
@@ -1,0 +1,140 @@
+using Beamable.Common.Api;
+using Beamable.Common.Inventory;
+using Beamable.Tests.Runtime;
+using NUnit.Framework;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine.TestTools;
+
+namespace Beamable.Server.Tests.Runtime
+{
+	/// <summary>
+	/// Document the non-null guarantees that <see cref="MicroserviceClientHelper.DeserializeResult{T}"/>
+	/// provided while it was backed by JsonUtility.FromJson (SDK 5.0.x and earlier).
+	///
+	/// A C# microservice serializes response DTOs with Newtonsoft using default NullValueHandling,
+	/// so an unassigned reference-type field arrives at the client as an explicit JSON null member,
+	/// for example {"name": null}. JsonUtility deserialized such a member to its Unity-serialization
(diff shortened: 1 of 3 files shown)


allister-beamable:
Superseded by #4755, which fixes the ObjectMapper null handling and carries equivalent regression coverage (explicit-null and absent members across string/array/List/Dictionary/nested types, populated round trip, plus a self-referential depth guard). The Unity-runtime suite on this branch exercises the same contract through MicroserviceClientHelper.DeserializeResult end to end and passes against the #4755 behavior; the branch stays available if that layer of coverage is wanted as a follow-up.

PR #4755, opened 2026-07-13
Fix: Added support to serialize/deserialize a null value we are defaulting it instead to prevent null reference errors.

### LINKED ISSUE #4756: Unity SDK 5.1 microservice client deserializes null and absent reference fields to null instead of empty defaults
Unity SDK 5.1.0 changed the object fallback of `MicroserviceClientHelper.DeserializeResult<T>` from `JsonUtility.FromJson<T>` to SmallerJSON `Json.Deserialize<T>` (74dcc190e).

**Expected** (5.0.x behavior): response DTO reference fields arriving as explicit JSON nulls deserialize to non-null defaults (`""`, empty array/list, default-constructed nested object); absent `List`, nested object, and ContentRef members are also default-constructed. The default microservice response serializer emits unassigned reference fields as explicit nulls (identically on 4.3.8 and current), so every object response takes this shape.

**Actual**: those members now deserialize to `null`, causing NullReferenceExceptions in game code written against the 5.0.x guarantees. Affects top-level object responses; the list path is unchanged.

Fix + regression tests: #4755. Additional Unity-runtime tests demonstrating the regression (red on 5.1.x, green at unity-sdk-5.0.1): #4754 (closed, branch `microservice-client-null-defaults-tests`).

DiasAtBeamable:
Fixes #4756

### DIFF
--- a/cli/beamable.common/Runtime/SmallerJSON/SmallerJSON.cs
+++ b/cli/beamable.common/Runtime/SmallerJSON/SmallerJSON.cs
@@ -1091,11 +1091,11 @@ namespace Beamable.Serialization.SmallerJSON
 			{
 				if (targetType == null) throw new ArgumentNullException(nameof(targetType));
 
-				// Null handling
+				// Null handling — mirror Unity JsonUtility, which never leaves reference fields null
+				// (missing string => "", missing array/list => empty, missing [Serializable] class => instance).
 				if (node == null)
 				{
-					targetType = Nullable.GetUnderlyingType(targetType) ?? targetType;
-					return targetType.IsValueType ? Activator.CreateInstance(targetType) : null;
+					return CreateJsonUtilityDefault(targetType, 0);
 				}
 
 				// Unwrap nullable
@@ -1185,7 +1185,13 @@ namespace Beamable.Serialization.SmallerJSON
 						var canBeSerialized = (hasSerializeField || !field.IsNotSerialized);
 						if (!canBeSerialized) continue;
 
-						if (!objNode.TryGetValue(field.Name, out var rawValue)) continue;
+						if (!objNode.TryGetValue(field.Name, out var rawValue))
+						{
+							// Field omitted from JSON: default to a JsonUtility-parity non-null value
+							// so downstream code keeps its historical non-nullable guarantees.
+							field.SetValue(instance, CreateJsonUtilityDefault(field.FieldType, 0));
+							continue;
+						}
 
 						var converted = ConvertToType(rawValue, field.FieldType);
 						field.SetValue(instance, converted);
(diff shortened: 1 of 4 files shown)


cli/beamable.common/Runtime/SmallerJSON/SmallerJSON.cs
```diff
@@ -1091,11 +1091,11 @@ public static object ConvertToType(object node, Type targetType)
 			{
 				if (targetType == null) throw new ArgumentNullException(nameof(targetType));
 
-				// Null handling
+				// Null handling — mirror Unity JsonUtility, which never leaves reference fields null
+				// (missing string => "", missing array/list => empty, missing [Serializable] class => instance).
```
allister-beamable, line 1095: Actually, the old behavior was missing string => null. Putting in a blank string is arguably better, but we should call it out in the changelog, and be accurate about what it used to do.

cli/beamable.common/Runtime/SmallerJSON/SmallerJSON.cs
```diff
@@ -1204,6 +1210,53 @@ public static object ConvertToType(object node, Type targetType)
...
 			}
 
+			// Guards against StackOverflow on self-referential [Serializable] graphs. Unity's own
+			// serializer stops nesting custom classes at a similar depth.
+			private const int MaxDefaultDepth = 10;
+
+			/// <summary>
+			/// Produces the same non-null default Unity's JsonUtility.FromJson would materialize for a
+			/// field that is absent (or explicitly null) in the JSON:
+			/// string => "", array => empty array, List/Dictionary => empty collection,
```
allister-beamable, line 1220: This needs the same accuracy treatment as line 1095: JsonUtility's behavior is to put `null` when a string field is absent.

allister-beamable: (however, it _should not ever happen_ that a `string` field declared in a C#MS return type gets actually omitted from the JSON: the NewtonSoft stuff on that end will faithfully preserve a `null` (which is the only idiom for expressing "this string is not present" unless one uses `OptionalString` in one's data structures)

allister-beamable (approved):
Good to go. The comments are ever so slightly inaccurate, but only about behavior that will never normally get hit in real Beamable C# Microservice <-> Beamable Unity SDK usage: JsonUtility.FromJson supplies `null` when a string field is truly absent, _BUT_ no string field should ever be truly absent from the JSON emitted by a C#MS.

I think it is worth keeping the absent -> `""` behavior because:
- maybe there will be Microservices that aren't official C#MS, and maybe their JSON-emission behavior will allow for absent fields
- using `""` as the value is the least surprising thing, given how JsonUtility.FromJson handles `null`s in string fields

Ship it!

cli/beamable.common/Runtime/SmallerJSON/SmallerJSON.cs
allister-beamable: Fun facts from code spelunking around `OptionalString`:
- we use it *a lot* in auto-generated code
- it also shows up in Content types that should have checkbox-enabled string entry fields

Bottom line: we do not normally expect game developer code to use `OptionalString` unless the use case very specifically demands optionality in addition to nullability, but it is there if anyone needs it.

allister-beamable:
Is it possible that this would affect Content serialization? Like client code relying on absent-field or null-field behavior? Is it even possible for a piece of Content to serialize with missing fields? We should have some unit tests that cover that even if the system is already perfectly safe.

allister-beamable:
Going to merge now to get an RC built; then @Leinnan will write content-oriented tests, if any new ones are needed.

PR #4763, opened 2026-07-17
Add publish history to the Content Manager #4509

### LINKED ISSUE #4509: [Unity] Add Content History window to Content Manager
Pedro is adding content history support to the CLI.

mhijaziB:
Fixes #4509 

# Description

Adds a new Content History tab for inspecting published content changes without leaving the editor.

- Browse publishes in time-based groups, including older monthly history.
- Search by manifest or content ID to narrow the audit trail.
- See publish author, changed-content count, and UTC ISO 8601 timestamps.
- Inspect a publish’s changelist and view historical content in a read-only preview.
- Restore selected historical content when needed, with loading and failure states kept separate from the history list.
- Includes fixes for search reset behavior, manifest ID copying, and history-stream resilience.

Tests: Unity EditMode content-history suite — 68/68 passing.

# Checklist

# Notes

### DIFF
--- a/client/Packages/com.beamable/CHANGELOG.md
+++ b/client/Packages/com.beamable/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleasd
+
+- Added Content Manager publish history with diff inspection, pagination, preview, and local file restoration.
+
 ## [5.1.1] - 2026-06-29
 
 ### Added
--- a/client/Packages/com.beamable/Editor/BeamCli/UI/LogUtil/LogViewUtil.cs
+++ b/client/Packages/com.beamable/Editor/BeamCli/UI/LogUtil/LogViewUtil.cs
@@ -238,7 +238,7 @@ namespace Beamable.Editor.BeamCli.UI.LogHelpers
 		public Action onEndCheck;
 	}
 
-	internal static class SearchBarClearInteraction
+	public static class SearchBarClearInteraction
 	{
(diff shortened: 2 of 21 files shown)


Leinnan (commented):
There is one thing I would change, I would replace all usages of `internal` with `public`.

allister-beamable:
> There is one thing I would change, I would replace all usages of `internal` with `public`.

+1; `internal` creates artificial barriers and friction for not much value. If/when anyone needs to modify or override this system's behavior, it will be easier to change it if things are `public`. Let us achieve encapsulation and separation of concerns by means of discipline rather than keywords.

allister-beamable:
> > There is one thing I would change, I would replace all usages of `internal` with `public`.
> 
> +1; `internal` creates artificial barriers and friction for not much value. If/when anyone needs to modify or override this system's behavior, it will be easier to change it if things are `public`. Let us achieve encapsulation and separation of concerns by means of discipline rather than keywords.

Devil's advocate: `internal` has some value when it is truly useful for signaling interface contracts: stuff marked `public` is stuff we implicitly promise will not breakingly change except when the major version bumps, and fills up IntelliSense.

Still, another argument against `internal`: nothing else in that file is marked internal; all the other classes in that namespace in that file are `public`. (there are some members within those classes that are implicitly private that ought to have an explicit `private` modifier, but that is a different issue entirely)

Anyhow, all this to say, "please make this one `public`, but we might have other places in the SDK that merit `internal`" 😸

mhijaziB:
@allister-beamable @Leinnan

Addressed the visibility feedback in b78a0ada6.

The Content History helpers and `CancelContentHistoryRequests()` are now public. This removes the editor-assembly boundary for consumers who need to extend or alter the flow, while keeping the implementation behavior unchanged.

Added a regression test to preserve that contract. Unity EditMode content-history suite: 69/69 passing.

allister-beamable (approved):
With revisions, I think this is good to go! Let's QA it as part of 6.0.0 RC.

PR #4764, opened 2026-07-20
Unity SDK version headers should be sent by default

### LINKED ISSUE #4758: Unity SDK version headers should be sent by default
## Summary

The Beamable Unity SDK can attach version headers to its API requests, but the code that does so is compiled out unless a define symbol is set, so no shipping build sends them by default. As a result there is no server-side signal for which SDK version a client is running.

This came up during a support investigation where we tried to determine which Beamable Unity SDK version a customer's realms were on, and found no reliable request-level telemetry for it.

## Current behavior (verified in-tree)

`AddVersionHeaders` in `client/Packages/com.beamable/Runtime/Core/Platform/SDK/PlatformRequester.cs` (~L557-566, called from the main request path ~L639) is doubly gated:

```csharp
[Conditional("BEAMABLE_ENABLE_VERSION_HEADERS")]
protected void AddVersionHeaders(Dictionary<string, string> headers)
{
#if !BEAMABLE_DISABLE_VERSION_HEADERS
    headers[HEADER_BEAMABLE_VERSION]    = _beamableVersion.ToString();
    headers[HEADER_APPLICATION_VERSION] = Application.version;
    headers[HEADER_UNITY_VERSION]       = Application.unityVersion;
    headers[HEADER_ENGINE_TYPE]         = $"Unity-{Application.platform}";
#endif
}
```

`BEAMABLE_ENABLE_VERSION_HEADERS` is not defined in a default build, so the method is a no-op and none of these headers are sent:

- `X-KS-BEAM-SDK-VERSION` - SDK version
- `X-KS-USER-AGENT` - engine/platform (e.g. `Unity-Android`)
- `X-KS-USER-AGENT-VERSION` - Unity version
- `X-KS-GAME-VERSION` - developer `Application.version`

(Header name constants: `client/Packages/com.beamable/Common/Runtime/Constants/Implementations/RequesterConstants.cs`.)

## Expected

A default build sends the version headers, so the backend can observe which SDK version each client is running.

## Proposed change

Send the version headers by default: invert the gating so they are on out of the box, and keep `BEAMABLE_DISABLE_VERSION_HEADERS` as the opt-out for anyone who needs to suppress them (request size, privacy).

## Related

Recording these values server-side is tracked separately for the Scala stack and the C# stack (linked below). This SDK change is the prerequisite - without it there is nothing for the backend to record.

## Related issues

- beamable/BeamableProduct#4758 - send version headers by default (client prerequisite)
- beamable/BeamableBackend#907 - record SDK version at the Scala gateway
- beamable/BeamableAPI#653 - record SDK version in the C# API stack

Leinnan:
Fixes #4758

### DIFF
--- a/client/Packages/com.beamable/Runtime/Core/Platform/SDK/PlatformRequester.cs
+++ b/client/Packages/com.beamable/Runtime/Core/Platform/SDK/PlatformRequester.cs
@@ -1,9 +1,3 @@
-#if !BEAMABLE_DISABLE_VERSION_HEADERS
-#define BEAMABLE_ENABLE_VERSION_HEADERS
-#else
-#undef BEAMABLE_ENABLE_VERSION_HEADERS
-#endif
-
 using Beamable.Api.Analytics;
 using Beamable.Api.Caches;
 using Beamable.Api.Connectivity;
@@ -64,7 +58,7 @@ namespace Beamable.Api
 			// if the routing map is empty, do not include the header
 			if (string.IsNullOrEmpty(self.RoutingMap))
 				return headers;
-			
+
 			headers[Constants.Requester.HEADER_ROUTINGKEY] = self.RoutingMap;
 			return headers;
 		}
@@ -385,7 +379,7 @@ namespace Beamable.Api
 				contentType = "application/json";
 			}
 
-			var safetyClone = new SDKRequesterOptions<T>(req); // TODO: since its a struct, do we need to do this? 
+			var safetyClone = new SDKRequesterOptions<T>(req); // TODO: since its a struct, do we need to do this?
 			return MakeRequestWithTokenRefresh<T>(contentType, bodyBytes, safetyClone);
 		}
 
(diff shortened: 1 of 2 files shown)


client/Packages/com.beamable/Runtime/Core/Platform/SDK/PlatformRequester.cs
```diff
@@ -548,13 +542,12 @@ protected Promise<T> MakeRequest<T>(
...
 				return false;
 			});
 
 		}
-		
-		[Conditional("BEAMABLE_ENABLE_VERSION_HEADERS")]
+
 		protected void AddVersionHeaders(Dictionary<string, string> headers)
 		{
 #if !BEAMABLE_DISABLE_VERSION_HEADERS
```
allister-beamable, line 553: So with the `BEAMABLE_DISABLE_VERSION_HEADERS` preprocessor symbol set, one could still suppress these headers, but the default is for that symbol to be absent, right? Is it exposed in Project Settings as a specific setting, or would anyone who wants to suppress the headers need to put it in the symbol defines field by name?

(aside, it seems quite weird that we had two opposite-direction settings to govern this behavior, and that one of them was automatically set by the other in the ifdefs at the top of the file)

allister-beamable (approved):
Good. I think we should have been emitting version headers by default all the time since 0.0.0, so this brings us to where we should have been all along.

PR #4767, opened 2026-07-22
Allow opting out of the init-time commerce SKU request #4595

### LINKED ISSUE #4595: Allow opting out of the init-time commerce (GetSKUs) API call
## Summary

`BeamContext` initialization triggers `GET /basic/commerce/skus` on every player session. Games that do not use the commerce subsystem still pay for this call in their pay-as-you-grow API quota, with no first-class way to suppress it.

## Current behavior

`BeamContext.SetupPurchaser` resolves `IBeamablePurchaser` and calls `Initialize` on it. The SDK self-registers `UnityBeamablePurchaser`, whose `Initialize` calls `PaymentService.GetSKUs()` *before* checking whether any SKUs are configured. So even with zero SKUs configured, the API call fires once per session.

## Workaround that exists today

Customers can replace the SDK's purchaser registration with the SDK-supplied `DummyPurchaser` via a `[BeamContextSystem]`:

```csharp
[BeamContextSystem]
public static class DisableCommerceInit
{
    [RegisterBeamableDependencies(order: 1000)]
    public static void RegisterServices(IDependencyBuilder builder)
    {
        builder.RemoveIfExists<IBeamablePurchaser>();
        builder.AddSingleton<IBeamablePurchaser, DummyPurchaser>();
    }
}
```

This works, but it is undocumented and requires customers to know about `DummyPurchaser`, the `RegisterBeamableDependencies` order field, and the difference between "no SKUs configured" and "purchaser unregistered."

## Proposed

A first-class configuration toggle - either on `BeamContextConfiguration` or via a project-settings flag - to suppress this init-time fetch:

- `skipCommerceInit` (or equivalent): when true, register `DummyPurchaser` automatically; do not call `GetSKUs`.

## Why this matters

Per-customer the cost is modest (rough order: pay-as-you-grow customers running ~1M+ sessions/month pay tens of dollars/month for this call), but the pattern is broad: any game that initializes `BeamContext` without using commerce pays for traffic it cannot control without DI tricks. The fix is a clean, low-risk SDK addition with a known opt-out code path already in place.

## Related

The mail portion of this issue was split out into #4623, since the dominant driver of init-time mail traffic turned out to be the social/friends system rather than direct mail use.

mhijaziB:
Fixes #4595 

## Summary

- Adds `CoreConfiguration.SkipCommerceInitialization`, defaulting to `false`.
- When enabled, Beamable registers `DummyPurchaser` and skips the init-time `GET /basic/commerce/skus` request.
- We apply the override from `Unity.Beamable` because it already references `Unity.Beamable.Purchasing`. Referencing `CoreConfiguration` from the purchasing assembly would create a circular assembly dependency.

## Testing

- Added Play Mode coverage for both the default and opt-out initialization paths.
- Manual A/B testing, when config is enabled `GetSkus` doesn't get called and the runtime purchaser becomes the `DummyPurchaser`.

##Screenshot
image

# Checklist

# Notes

### DIFF
--- a/client/Packages/com.beamable/CHANGELOG.md
+++ b/client/Packages/com.beamable/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added Content Manager publish history with diff inspection, pagination, preview, and local file restoration.
 - Add support for the version 5 of the Unity IAP package
+- Added `CoreConfiguration.SkipCommerceInitialization` to opt out of automatic purchaser initialization and its commerce SKU request. [4595](https://github.com/beamable/BeamableProduct/issues/4595)
 
 ### Fixed
 
--- a/client/Packages/com.beamable/CHANGELOG.md
+++ b/client/Packages/com.beamable/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added Content Manager publish history with diff inspection, pagination, preview, and local file restoration.
 - Add support for the version 5 of the Unity IAP package
-- Added `CoreConfiguration.SkipCommerceInitialization` to opt out of automatic purchaser initialization and its commerce SKU request. [4595](https://github.com/beamable/BeamableProduct/issues/4595)
+- Added `CoreConfiguration.SkipCommerceInitialization` to opt out of automatic purchaser initialization and its commerce SKU request.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


client/Packages/com.beamable/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added Content Manager publish history with diff inspection, pagination, preview, and local file restoration.
 - Add support for the version 5 of the Unity IAP package
+- Added `CoreConfiguration.SkipCommerceInitialization` to opt out of automatic purchaser initialization and its commerce SKU request. [4595](https://github.com/beamable/BeamableProduct/issues/4595)
```
allister-beamable: Historically we have not referenced GitHub Issues by number or link in our changelog. I think omitting the links makes for a cleaner, more to-the-point changelog, and those who are truly fascinated by GitHub Issues can always navigate to issues directly. 😸

PR #4785, opened 2026-07-28
Issue/zone support

cdhanna:


cli/beamable.common/beamable.common.csproj
```diff
@@ -47,6 +47,12 @@
         <PackageReference Include="Newtonsoft.Json" Version="13.0.3"/>
     </ItemGroup>
     
+    <ItemGroup>
```
gabrielbeamable: This shouldn't be added, otherwise it won't compile some necessary endpoints for the bundles work

gabrielbeamable (changes requested):

Issue #4789, opened 2026-07-30
Unity IAP 5.4.2 support


Leinnan:
## Describe the Bug 

Unity IAP version 5.4.2 introduced new field for the `IOrderInfo` interface. This breaks `SteamOrderInfo`.

### Steps to reproduce the behavior:

1. Install Beamable SDK 6.0.0
2. Install Unity IAP 5.4.2
3. Notice compilation error.

### Expected Behavior

It compiles correctly

### Actual Behavior

Compilation Error

PR #4790, opened 2026-07-30
Unity IAP 5.4.2 support

Leinnan:
Implements #4789 . Implements missing field for the `SteamOrderInfo`.


allister-beamable (approved):
Ship it! My understanding is that the 5.4.1 to 5.4.2 difference affects OrderInfo, and so within the context of Beamable integration, that only pertains to Steam purchasing because our Apple and Google integrations do not rely on OrderInfo.

PR #4792, opened 2026-08-03
Add build-scoped Player Social mail suppression with runtime opt-in

### LINKED ISSUE #4623: Allow opting out of the mail service to avoid social/friends-triggered mail API calls at startup
## Summary

Games that do not knowingly use mail still see `POST /object/mail/{userId}/search` traffic on the API usage dashboard, roughly 1:1 with login. The driver is the **social/friends** system, not a mail UI: friend invitations ride on the mail channel, so `PlayerSocial` reads mail on construction and on every mailbox update.

## Current behavior

The `PlayerSocial` constructor (`Runtime/Player/PlayerFriends.cs:64-86`) subscribes to `MAILBOX.UPDATE` and schedules `RefreshMail()` on ready. `RefreshMail()` (`PlayerFriends.cs:104-111`) calls `_mailApi.GetMail("SOCIAL.FRIEND.INVITE")`, which `AbsMailApi.GetMail` (`Common/Runtime/Api/Mail/AbsMailApi.cs:44-57`) sends as `POST /object/mail/{userId}/search`. Because `PlayerSocial` is lazily resolved via `BeamContext.Social` (`Runtime/BeamContext.cs:232`), any game that touches Friends generates a mail search on startup and again on every `MAILBOX.UPDATE` — with no mail code of its own.

Customers in this situation are surprised to see mail hits on the dashboard.

## Proposed

Friends and mail cannot currently be decoupled: invitations are delivered over the mail channel, so the mail search is load-bearing for the invitations feature. The realistic option is a `BeamContextConfiguration` / project-settings toggle that forcibly suppresses mail usage for games that do not use player-facing mail or friend invitations.

**Important caveat:** such a toggle would break the invitations feature of Friends & Social, since invites depend on the mail fetch. It is an opt-out for games that want neither mail nor invites, not a way to keep invitations while dropping the traffic.

At minimum, document the social/friends → mail coupling so the dashboard traffic is not surprising.

Notification is intentionally **not** in scope: the Notifications system underpins every Beamable subscription, so disabling its per-login setup would break core SDK behavior.

## Why this matters

Per-customer the cost is modest (rough order: pay-as-you-grow customers running ~1M+ sessions/month pay tens of dollars/month for these calls), but the coupling is non-obvious: a game can use only Friends and still pay for mail traffic it never asked for, with no documented opt-out.

## Related

Split out from #4595, which now covers the commerce (`GetSKUs`) init-time call only.

mhijaziB:
Fixes #4623 

## What changed
Player Social automatically searches Mail for friend invitations when it initializes and subscribes to future mailbox updates. Projects that do not use Friends still pay for those Mail requests.

This PR adds two levels of control:
- A Unity Project Settings flag, `SuppressAutomaticMailUpdatesFromPlayerSocial`, that disables the automatic Mail search and `MAILBOX.UPDATE` subscription.
- A runtime method, `EnableFriendInvitationMailChecks()`, that enables those checks for a specific `PlayerSocial/player` context when the game actually needs friend invitations.
-
Default behavior is unchanged. Projects must explicitly enable the suppression setting.

## Behavior when suppressed
Existing friends and other Social functionality remain available, but sending or accepting friend invitations throws an `InvalidOperationException` until the game calls:

```
await context.Social.EnableFriendInvitationMailChecks();
```

Calling this method performs an immediate Mail search and subscribes that player context to future mailbox updates. Enabling one `PlayerSocial` instance does not enable another.

This suppression only affects the Mail usage owned by Player Social. Other Mail services continue to work normally.

## Testing
Added focused Play Mode coverage for:
- Default startup behavior still searching Mail.
- Suppressed startup behavior making no Mail search.
- Sending and accepting invitations being rejected while suppressed.
- Enabling invitation Mail checks mid-session.
- Repeated enable calls remaining idempotent.
- Mailbox notifications triggering subsequent Mail searches.
- Sending invitations after runtime enablement.
- The override remaining scoped to the selected `PlayerSocial` instance.
- All 9 focused Play Mode tests pass, and the Unity C# test project builds successfully.

# Checklist

# Notes


client/Packages/com.beamable/Runtime/Player/PlayerFriends.cs
```diff
@@ -341,10 +387,41 @@ public async Promise ImportFacebookFriends(string thirdPartyAuthToken)
...
+		/// This performs an immediate Mail search and subscribes this player
+		/// context to future mailbox updates.
+		/// </remarks>
+		public async Promise EnableFriendInvitationMailChecks()
+		{
+			if (_friendInvitationMailChecksEnabled) return;
+			await OnReady;
+
+			// Recheck because another caller may have enabled it while awaiting.
+			if (_friendInvitationMailChecksEnabled) return;
```
allister-beamable: The double early-out feels kind of redundant. What is the cost of only doing the check after `await OnReady`? If it is negligible, I would say get rid of the first check on line 400.

client/Packages/com.beamable/Runtime/Player/PlayerFriends.cs
```diff
@@ -230,8 +264,14 @@ public override int GetBroadcastChecksum()
 		/// After the resultant <see cref="Promise"/> completes, the <see cref="SentInvites"/> list will contain an invite for the given <see cref="playerId"/>
 		/// </summary>
 		/// <param name="playerId">the player id of the player to invite to become friends</param>
+		/// <exception cref="InvalidOperationException">Thrown when friend invitation Mail checks are disabled for this player context.</exception>
 		public async Promise Invite(long playerId)
 		{
+			if (!_friendInvitationMailChecksEnabled)
+			{
+				throw new InvalidOperationException(MAIL_SUPPRESSION_MESSAGE);
+			}
```
allister-beamable, line 273: Could this (and other `_friendInvitationMailChecksEnabled` checks, _specifically on public methods_) do an ad hoc initialization of the mail stuff? Thinking out loud: the philosophical razor is "explicit" vs "implicit" and the old wisdom is that explicit is better than implicit. (and what you have here falls on the "explicit" side)

One argument for implicit is that it _might_ be the path of least surprise: if, as a game developer, I am calling `Invite` or `AcceptInviteFrom`, maybe I expect to implicitly waive my previous suppression of social-based mail stuff? I think I am talking myself out of the "implicit" argument here: it would add complexity for something that at best removes a tiny bit of friction from one path and at worst surprises anyone who _did_ expect the explicit mode.

Nothing to see here, move along. (though I think the comment worth keeping as a design philosophy signpost)

allister-beamable (approved):
The one change I would like to see, if it is reasonable, is de-redundant-izing the field check on lines 400 and 404 of PlayerFriends.cs

client/Packages/com.beamable/Runtime/Player/PlayerFriends.cs
mhijaziB: Usually for mail service, this `OnReady` check is not that expensive compared for example to content or accounts. I will remove it

client/Packages/com.beamable/Runtime/Player/PlayerFriends.cs
mhijaziB: Agreed. Keeping this explicit preserves the meaning of the suppression setting: Invite and AcceptInviteFrom will not silently re-enable Mail traffic. Developers who need those flows can explicitly enable them for that player context through EnableFriendInvitationMailChecks() first. I’ll leave the public method guards as-is.

mhijaziB:
Waiting for checks, and then will merge

PR #4795, opened 2026-08-06
Allow tuples in microservice callables outside Unreal compatibility

### LINKED ISSUE #4582: Beamable.Microservice.SourceGen no longer allows tuples
A BPC customer was using client callables with tuple returns, 

```csharp
public (int, int) Test(){
 return (1,3);
}
```

This works fine in the serialization for Unity. The source analyzer reports an error saying that the type must have the `[Serializable]` attribute, but tuples _cannot_ have that attribute. The BPC customer disabled the analyzer to work around the issue, but that is not optimal. 

Given that Unreal does not support this, this easiest fix is to gate tuple support under the existing "Is this unreal" property.

mhijaziB:
# Ticket

Closes #4582

# Brief Description

`ValueTuple` return values already serialize correctly for Unity clients, but the microservice source analyzer treated the tuple container as a normal struct and required `[Serializable]`. Tuples cannot carry that attribute, so developers had to disable the analyzer even though the runtime path worked.

This PR allows tuple containers when Unreal Blueprint compatibility is disabled. We still validate every tuple element, so unsupported nested types continue to report diagnostics.

We keep the platform boundary explicit:

- Unity/C# client callables can use tuples in return types, parameters, and serializable response fields.
- Unreal Blueprint-compatible callables still reject tuples.
- `[BeamGenerateSchema]` fields still reject tuples because schema generation requires attribute-bearing types.

The branch also corrects the recursive analyzer argument order so nested fields receive the intended Unreal-compatibility and `BeamGenerateSchema` flags.

Client-side coverage confirms that generated methods preserve tuple parameter and return types, request serialization round-trips tuple parameters, and response deserialization round-trips tuple returns.

# Developer Impact

Developers using Unity/C# can keep the source analyzer enabled when a client callable uses tuples. Unreal projects retain the existing diagnostic instead of generating an unsupported Blueprint type.

# Verification

- `dotnet test templates/SourceGen/MicroserviceSourceGen.Tests/MicroserviceSourceGen.Tests.csproj --no-build --no-restore --nologo --verbosity quiet`
  - 36 passed, 0 failed
- `dotnet test cli/tests/tests.csproj --no-restore --nologo --verbosity minimal --filter "FullyQualifiedName~ValueTupleMicroserviceClientTests"`
  - 3 passed, 0 failed
- Mutation check: restoring the previous swapped recursive arguments makes the nested Unreal tuple-field test fail.

# Checklist

- [x] Added focused analyzer, code-generation, request-serialization, and response-deserialization coverage.
- [x] Changelog entry is not required; this fixes analyzer behavior without adding or changing a public API.
- [x] Public API documentation is not required; no public methods or interfaces were added.

# Notes

This does not introduce a new compatibility promise for Unreal or `[BeamGenerateSchema]`. Those paths remain intentionally unsupported for tuples.

### DIFF
--- a/cli/tests/Unity/ValueTupleMicroserviceClientTests.cs
+++ b/cli/tests/Unity/ValueTupleMicroserviceClientTests.cs
@@ -19,7 +19,7 @@ namespace tests.Unity;
 public class ValueTupleMicroserviceClientTests
 {
 	[Test]
-	public void ClientGenerator_PreservesValueTupleCallableParameterType()
+	public void ClientGenerator_PreservesValueTupleCallableTypes()
 	{
 		InitializeLogging();
 		var document = GenerateServiceDocument();
@@ -34,6 +34,8 @@ public class ValueTupleMicroserviceClientTests
 				Does.Contain("System.ValueTuple<int, int> value"));
 			Assert.That(generatedClient,
 				Does.Contain("serializedFields.Add(\"value\", raw_value)"));
+			Assert.That(generatedClient,
+				Does.Contain("Beamable.Common.Promise<System.ValueTuple<int, int>> GetTuple()"));
 		}
 		finally
 		{
@@ -94,5 +96,8 @@ public class ValueTupleMicroserviceClientTests
 	{
 		[ClientCallable]
 		public int AddTuple((int, int) value) => value.Item1 + value.Item2;
+
+		[ClientCallable]
+		public (int, int) GetTuple() => (1, 3);
 	}
 }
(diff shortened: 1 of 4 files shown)


DiasAtBeamable (approved):
LGTM

Leinnan (approved):
Good work 🐗

allister-beamable (approved):
Just to make sure I understand the fix:
- the JSON serializer we use in Unity has always had the capability to ser/deser tuples
- validation rules forbade tuples broadly
  - (edit to add: and we introduced those validation rules between Beamable Unity SDK v4.x and v5.x)
- this fix relaxes the validation rules when in a non-Unreal-blueprints context
Is that accurate?

Issue #4728, opened 2026-06-29
BeamWebCommand blocks the Editor main thread on StreamReader.EndOfStream (CLI stream read)


allister-beamable:
### What

`BeamWebCommand.RunHttpRequestOnce` reads the CLI local-server response stream with a `while (!reader.EndOfStream)` loop. `StreamReader.EndOfStream` is a **synchronous, blocking** property — it forces a read of the underlying network stream and blocks the calling thread until the next byte arrives or the connection closes. Because the surrounding `async` continuations resume on Unity's main-thread `SynchronizationContext`, that blocking read runs **on the Editor main thread**, freezing the entire Editor (repaints, input, text entry) until the next CLI stream chunk arrives.

### Expected

Reading the CLI stream never blocks the Editor main thread; the loop yields between chunks so the Editor stays responsive while a CLI command streams.

### Actual

A single captured Editor frame spent **756ms** parked in `Monitor.Wait`, with this call chain (Profiler, standalone process, Deep Profile):

```
EditorLoop → Application.Tick → Application.UpdateScene → Update.ScriptRunDelayedTasks
  → UnitySynchronizationContext.ExecuteTasks → …<RunHttpRequestOnce>d__14.MoveNext()
  → StreamReader.get_EndOfStream → StreamReader.ReadBuffer → HttpContentReadStream.Read
  → Task.SpinThenBlockingWait → ManualResetEventSlim.Wait → Monitor.ObjWait → Monitor.Monitor_wait() [756.68ms self]
```

The loop body itself is already async (`ReadLineAsync()` + `Task.WhenAny`), so only the `EndOfStream` loop **condition** blocks — it defeats the otherwise-async pump.

### Location

`client/Packages/com.beamable/Editor/BeamCli/BeamWebCommand.cs` — the `while (!reader.EndOfStream)` loop in `RunHttpRequestOnce` (around line 473).

### Fix direction

Drive the loop off the async read result instead of the blocking property: loop on `await reader.ReadLineAsync()` and break when it returns `null` (end of stream), rather than polling `reader.EndOfStream`. That keeps every stream read off the main thread.

### Environment

Beamable Unity SDK 5.1.1-PREVIEW.RC1 · CLI 7.2.1-PREVIEW.RC1 · Unity 6000.2.15f1 · macOS. Caught via an auto-freeze-on-spike Editor profiling harness.

allister-beamable:
[...]

PR #4800, opened 2026-08-12
Cache content validation results in the Unity Inspector

### LINKED ISSUE #4727: Content inspector re-validates every field on each repaint (uncached), making typing syrupy
### What

The content Inspector re-runs full field validation on **every IMGUI repaint** — and twice per repaint — with no caching. While typing in any field, the whole inspector repaints per keystroke, so every validation-annotated field is re-validated 2× per keystroke. This makes editing feel sluggish ("syrupy").

### Expected

Validation results are cached and recomputed only when a field value actually changes, not on every repaint. Typing in a field does not re-validate the entire inspector.

### Actual

`ContentValidationPropertyDrawer` (`client/Packages/com.beamable/Editor/Modules/Content/ContentValidationPropertyDrawer.cs`) calls `attribute.Validate(...)` in **both** `OnGUI` (line ~189) and `GetPropertyHeight` (line ~52). IMGUI invokes both every repaint, so each annotated field is validated twice per repaint. Each pass also does reflection-based work that isn't cached:

- `ContentRefPropertyDrawer.GetTargetParentObjectOfProperty` / `GetTargetObjectOfProperty` — reflection walk of the property path, every pass.
- A fresh `ValidationFieldWrapper` + `ContentValidationArgs` per pass; `MustReferenceContent.Validate` → `ValidateId` resolves the field value and looks the id up against the validation context.
- The fallback `RefEditorGUI.DefaultPropertyField` does `Activator.CreateInstance(drawerType)` on every call.

### Profiler evidence

Standalone-process Profiler, Deep Profile, a repaint frame while the inspector was visible (CPU 17.5ms):

```
InspectorWindow.Repaint (15.7ms) → … → ContentObjectEditor.OnInspectorGUI (10.99ms)
  → DrawDefaultInspector → EditorGUILayout.PropertyField ×24 → PropertyHandler.OnGUI
  → PropertyDrawer.OnGUI → ContentValidationPropertyDrawer.OnGUI  [7.19ms / 8 calls]
```

~7ms of an ~11ms inspector redraw is the validation drawer.

### Magnitude / scope

~11ms per repaint is the validation drawer alone; the whole `EditorLoop` frame measured **17.36ms — already past the 16.6ms 60fps budget** — and this was on a high-end MacBook Pro, i.e. near best case. The validation cost scales with CPU, so on weaker hardware it is not sub-frame: it crosses from "syrupy" into dropped frames while typing. Because IMGUI repaints the whole inspector on every keystroke, every keystroke pays this in full.

The multi-hundred-millisecond freezes felt while editing are a **separate** bug: #4728 (main-thread block on a synchronous CLI stream read). This issue is the steady-state redraw cost; #4728 is the intermittent freeze.

### Fix direction

Cache validation results per (object, field) and recompute only on value change; avoid re-validating in `GetPropertyHeight` (compute height from a cached result). Cache the reflected target-object lookup and the resolved drawer type instead of recomputing per repaint.

### Environment

Beamable Unity SDK 5.1.1-PREVIEW.RC1 · CLI 7.2.1-PREVIEW.RC1 · Unity 6000.2.15f1 · macOS.

mhijaziB:
Fixes #4727 

## What changed
Content validation was running from both `GetPropertyHeight` and `OnGUI` during every Inspector event, including repaints, mouse movement, and clicks. This caused repeated validation work, excessive logging from validators, and unnecessary Editor overhead.

Now we route both callbacks through a shared validation-result cache. Cached results are invalidated when:

- The serialized property changes
- The content manifest changes
- Unity performs Undo/Redo
- The selected object changes
- Scripts reload or the Editor closes

The cache is bounded to prevent stale entries from accumulating.

## Verification

- Confirmed that selecting content still renders its Inspector correctly
- Confirmed that hovering and clicking without changing values no longer reruns validation
- Confirmed that editing a validated field invalidates the cached result
- Confirmed that validation messages and property height update after changes
- Built `Unity.Beamable.Editor.csproj` with 0 warnings and 0 errors

The delayed content-save hitch is a separate problem covered by #4728  and is intentionally outside this PR.

# Checklist

# Notes


Leinnan (approved):
lgtm

client/Packages/com.beamable/Editor/Modules/Content/ContentValidationPropertyDrawer.cs
```diff
@@ -18,48 +18,266 @@ public class ContentValidationPropertyDrawer : PropertyDrawer
 		private GUIStyle _lblStyle;
 		private const int WIDTH = 3;
 		private const int OFFSET = -10;
+		private const int MAX_VALIDATION_CACHE_ENTRIES = 512;
```
allister-beamable, line 21: Out of curiosity, how did you arrive at 512 for this? Just a safe number that is a pleasant power of 2? What size (in bytes) of on-disk or in-memory cache will this lead to, or is it highly variable based on actual existing content?

allister-beamable (approved):
Overall looks good.

client/Packages/com.beamable/Editor/Modules/Content/ContentValidationPropertyDrawer.cs
mhijaziB: It is more of safety and not a magic number per say. Power of 2 here has not benefits. 512 is reasonable enough as valid entries use around 100-200kbg on a 64 bit Editor. Invalid results use way more because they retain exception objects and message strings. So reaching 512 while inspecting normal content would be unusual.

PR #4801, opened 2026-08-13
Fix Unity Editor hitches during content saves

### LINKED ISSUE #4728: BeamWebCommand blocks the Editor main thread on StreamReader.EndOfStream (CLI stream read)

mhijaziB:
Fixes #4728 

## Summary

- Read streamed CLI responses until `ReadLineAsync()` returns `null` instead of polling synchronous `EndOfStream`.
- Preserve prompt cancellation and observe faults from abandoned pending reads.

## Verification

- Confirmed fast and slow Inspector typing remain smooth.
- Confirmed the previous main-thread hitch no longer appears in Unity profiling.
- Canceled an idle `content ps --watch` command; cancellation completed promptly, history reached Completed, and no delayed warnings or errors appeared.

# Checklist

# Notes

### DIFF
--- a/client/Packages/com.beamable/Editor/BeamCli/BeamWebCommand.cs
+++ b/client/Packages/com.beamable/Editor/BeamCli/BeamWebCommand.cs
@@ -470,7 +470,7 @@ namespace Beamable.Editor.BeamCli
 				using StreamReader reader = new StreamReader(streamToReadFrom);
 
 				Task<string> readTask = null;
-				while (!reader.EndOfStream)
+				while (true)
 				{
 					if (_cts.Token.IsCancellationRequested)
 					{
@@ -483,13 +483,27 @@ namespace Beamable.Editor.BeamCli
 					{
 						await Task.WhenAny(readTask, cancelTcs.Task);
 					}
+
 					if (_cts.Token.IsCancellationRequested)
 					{
+						_ = readTask.ContinueWith(completedReadTask => { _ = completedReadTask.Exception;},
+							CancellationToken.None,
+							TaskContinuationOptions.OnlyOnFaulted |
+							TaskContinuationOptions.ExecuteSynchronously,
+							TaskScheduler.Default);
 						break;
 					}
 
 					var line = await readTask;
-					if (string.IsNullOrEmpty(line)) continue; // TODO: what if the message contains a \n character?
+					if (line == null)
+					{
+						break;
+					}
+
+					if (line.Length == 0)
+					{
+						continue;
+					}
 
 					// remove life-cycle zero-width character
 					line = line.Replace("\u200b", "");
--- a/client/Packages/com.beamable/CHANGELOG.md
+++ b/client/Packages/com.beamable/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added build-scoped suppression and per-player runtime opt-in for automatic Player Social friend-invitation Mail checks.
 - Content Manifest caching between runs.
 
+### Fixed
+
+- Fixed Unity Editor hitches after editing content properties by avoiding synchronous end-of-stream checks while reading streamed CLI responses.
+
 ## [6.0.1] - 2026-07-31
 
 ### Changed


client/Packages/com.beamable/Editor/BeamCli/BeamWebCommand.cs
```diff
@@ -470,7 +470,7 @@ private async Promise RunHttpRequestOnce()
 				using StreamReader reader = new StreamReader(streamToReadFrom);
 
 				Task<string> readTask = null;
-				while (!reader.EndOfStream)
+				while (true)
```
allister-beamable, line 473: Why can't the condition stay as `!reader.EndOfStream`? Were we getting false reports of EOS that interrupted things inappropriately?

client/Packages/com.beamable/Editor/BeamCli/BeamWebCommand.cs
```diff
@@ -483,13 +483,27 @@ private async Promise RunHttpRequestOnce()
...
 						await Task.WhenAny(readTask, cancelTcs.Task);
 					}
+
 					if (_cts.Token.IsCancellationRequested)
 					{
+						_ = readTask.ContinueWith(completedReadTask => { _ = completedReadTask.Exception;},
+							CancellationToken.None,
+							TaskContinuationOptions.OnlyOnFaulted |
+							TaskContinuationOptions.ExecuteSynchronously,
+							TaskScheduler.Default);
```
allister-beamable, line 493: This is one last read upon cancellation? A "final hurrah" as it were, in hopes of getting the data if it can be gotten with a single call?

client/Packages/com.beamable/Editor/BeamCli/BeamWebCommand.cs
allister-beamable: Oh, I see based on the CHANGELOG entry: EndOfStream is a property that has side effects, not simply a field that can be read cheaply at every turn?

allister-beamable (approved):
Couple of "why is this the right solution?" questions; needn't drive any code changes or even new comments in the code -- as long as the conversation on this PR clarifies the "why?" that is good enough for me.

mhijaziB:
@allister-beamable 
Yes exactly, `SteamReader.EndOfSteam` can synchronously read from the stream to determine whether EOF has been reached. In Unity, this check was happening on the main thread and thus blocking it in short instances causing the inspector to hitch. 
Now `ReadLineAsync()` determine the state: `null` means EOF, while an empty string remains a valid blank line. 

The cancellation continuation is not a final read. 2021 doesn't support `ReadLineAsync()` . If cancellation wins, we exit immediately and dispose the response without waiting for the pending read since it could hang. isposal may later fault the abandoned read task, so the OnlyOnFaulted continuation observes that exception and prevents it from becoming an unobserved task exception. We intentionally do not process any late data; cancellation keeps priority, matching the previous behavior.

If you want me to add any comments in the code, please let me know

client/Packages/com.beamable/Editor/BeamCli/BeamWebCommand.cs
```diff
@@ -483,13 +483,27 @@ private async Promise RunHttpRequestOnce()
...
-					if (string.IsNullOrEmpty(line)) continue; // TODO: what if the message contains a \n character?
+					if (line == null)
+					{
+						break;
+					}
+
+					if (line.Length == 0)
+					{
+						continue;
+					}
```
Leinnan, line 506: Why not just use `string.IsNullOrWhitespace` to cover the `\n` case mentioned in original comment? Why this approach

Leinnan (approved):
lgtm

PR #4805, opened 2026-08-14
Fix logging unexpected curly-brace expressions

### LINKED ISSUE #4566: Logging blows up when LogError gets unexpected curly-brace expressions
## Describe the Bug

In `BeamableMicroService.ResolveCustomInitializationHook()`, the catch block that handles a failed `[InitializeServices]` method constructs its log message using C# string interpolation, then passes the result to `BeamableLogger.LogError` as a format string:

```csharp
BeamableLogger.LogError(
    $"Custom service initializer [{initializationMethod.DeclaringType?.FullName}.{initializationMethod.Name}] failed.\n" +
    $"{ex.Message}\n" +
    $"{{stacktrace}}",
    ex.StackTrace);
```

`$"{ex.Message}"` embeds the exception message directly into the format string via interpolation. If the exception message contains curly braces (e.g. from a JSON fragment, a C# format string, or any `{...}` text), the logging framework interprets them as additional placeholders. With only one argument supplied (`ex.StackTrace`), this causes a secondary `IndexOutOfRangeException` inside the logger, which propagates as:

> An error occurred while writing to logger(s). (Index (zero based) must be greater than or equal to zero and less than the size of the argument list.)

This secondary exception completely masks the original `[InitializeServices]` failure, making the root cause invisible to the developer.

## To Reproduce

Add an `[InitializeServices]` method that throws an exception whose message contains curly braces:

```csharp
[InitializeServices]
public static async Task ReproLoggerFormatStringBug(IServiceInitializer initializer)
{
    throw new Exception("Simulated init failure with {curly} {braces} in message");
}
```

Run the microservice locally (`beam project run`). The displayed error will be the logger's secondary crash, not the thrown exception.

### Expected Behavior

The original `[InitializeServices]` exception (message and stack trace) is logged clearly. The developer can see what actually went wrong.

### Actual Behavior

A secondary `IndexOutOfRangeException` from the logging framework is displayed instead. The original exception is lost. The error repeats multiple times in the output:

```
Service failed to provide services. An error occurred while writing to logger(s).
(Index (zero based) must be greater than or equal to zero and less than the size of the argument list.)
(Index (zero based) must be greater than or equal to zero and less than the size of the argument list.)
(Index (zero based) must be greater than or equal to zero and less than the size of the argument list.)
   at Microsoft.Extensions.Logging.Logger.ThrowLoggingError(List`1 exceptions)
   ...
   at Beamable.Common.BeamableLogger.LogError(String error, Object[] args)
   at Beamable.Server.BeamableMicroService.ResolveCustomInitializationHook()
   at Beamable.Server.BeamableMicroService.SetupWebsocket(IConnection socket, Boolean initContent)
```

## Suggested Fix

Pass `ex.Message` as a separate argument rather than interpolating it into the format string, and use a literal placeholder:

```csharp
BeamableLogger.LogError(
    "Custom service initializer [{typeName}.{methodName}] failed.\n{message}\n{stacktrace}",
    initializationMethod.DeclaringType?.FullName,
    initializationMethod.Name,
    ex.Message,
    ex.StackTrace);
```

## Metadata

- **Affected file:** `microservice/microservice/dbmicroservice/BeamableMicroService.cs`, `ResolveCustomInitializationHook()`
- **Observed in:** CLI 7.0.1 (`cli-7.0.1` tag)
- **Operating System:** macOS (arm64), confirmed also present in a Linux container environment

## Additional Context

This bug was discovered during a CLI 7.x upgrade support case. An `[InitializeServices]` hook threw an exception whose message contained curly braces, triggering the secondary logging crash. The misleading error message significantly delayed diagnosis - the real failure was only found after commenting out the init hook entirely.

Leinnan:
# Ticket

Fixes #4566 . 

# Brief Description

Introduces compile time detection of incorrect log calls. It also fixes the case with logging of the failed custom service initializer.


cli/cli/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Portal extension scanning no longer excludes sym linked package files
 - Fixed orphaned Unity .meta files left behind when cleaning generated Beamable source directories.
+- Fix logging unexpected curly-brace expressions  [4566](https://github.com/beamable/BeamableProduct/issues/4566).
```
mhijaziB: Please remove the reference to the ticket

microservice/beamable.tooling.common/Microservice/SafeLogTemplate.cs
```diff
@@ -0,0 +1,225 @@
...
+
+			if (!logger.IsEnabled(level))
+			{
+				return;
+			}
+
+			// The template is rendered here, up front, so that a broken one is caught while this frame is still
+			// on the stack. A try/catch around the write below would not be enough on its own: a queued logger
+			// defers formatting until its flush, and an otel exporter until its export.
+			if (!CanRender(template, args))
```
mhijaziB: Codex:
There is a chance that this still leaves an exception path inside the “safe” logger.
`CanRender` formats the real argument objects. If one of their `ToString()` implementations throws, `CanRender` returns false, then `Flatten` calls the same `ToString()` again outside any protection. Codex reproduced this with a structured `ILogger` that never formats the value itself, and `SafeLogTemplate.Write` still threw from `Flatten`.
This also means valid log arguments are formatted once during validation and then again by the actual provider.
Can we validate only the template syntax and argument count without formatting application objects, then make the fallback render each argument independently and safely? We should add a test with a throwing `ToString()` as well.

templates/MicroserviceSourceGen/MicroserviceSourceGen/Analyzers/LogTemplateAnalyzer.cs
```diff
@@ -0,0 +1,163 @@
...
+		return method.Name.EndsWith(COMPOSITE_FORMAT_METHOD_SUFFIX, StringComparison.Ordinal);
+	}
+
+	/// <summary>
+	/// Counts the values passed to a <c>params</c> parameter, or returns -1 when the caller passed an array
+	/// directly and the count cannot be known at compile time.
+	/// </summary>
+	private static int GetLogArgumentCount(IArgumentOperation logArguments)
+	{
+		if (logArguments.ArgumentKind != ArgumentKind.ParamArray)
```
mhijaziB: Codex: 
This skips caller-supplied or forwarded `object[]` arguments because Roslyn marks those as `ArgumentKind.Explicit`, not `ArgumentKind.ParamArray`.
Codex verified these cases:
```
BeamableLogger.LogError("two {a} {b}", new object[] { 1 });
BeamableLogger.LogError(message, args);
```
Both produced zero diagnostics, while the expanded `..., 1` call correctly produced `BEAM_LOG_0002`. Forwarding a params `object[]` is also a pattern we already use in `MicroserviceDebug`.
Can we inspect `IArrayCreationOperation` regardless of `ArgumentKind` and evaluate the nonconstant-template warning separately from whether the array length is known? I would also add analyzer tests for expanded arguments, explicit arrays, forwarded arrays, empty arrays, and both logging helper types.

microservice/beamable.tooling.common/LogUtil.cs
```diff
@@ -30,14 +30,24 @@ public void Flush(ILogger target)
...
+				{
+					target.Log(
+						logLevel: message.logLevel,
+						eventId: message.eventId,
+						state: message.state,
+						exception: message.exception,
+						formatter: message.formatter
+						);
+				}
+				catch (Exception ex)
```
mhijaziB, line 44: We still lose the original startup failure in this path.
If deferred formatting fails during Flush, this catch replaces the record with the formatting exception message. The original `message.exception`, template/state, event ID, and severity are all discarded. That is the diagnostic context we are specifically trying to preserve with this fix.
Can we emit an exception-safe representation of the original record while retaining the attached exception, then continue flushing the remaining queue? A focused test with one malformed exception-bearing record followed by a valid record would prove both behaviors.

templates/MicroserviceSourceGen/MicroserviceSourceGen/Diagnostics.cs
```diff
@@ -243,6 +244,41 @@ public static readonly DiagnosticDescriptor InvalidGenericTypeOnMicroservice
...
+		/// is read as part of the template. This is a warning rather than an error because the call is only
+		/// provably wrong once the interpolated text is known.
+		/// </summary>
+		public static readonly DiagnosticDescriptor NonConstantLogTemplate
+			= new(NON_CONSTANT_LOG_TEMPLATE_ID,
+				"Log message template must be a constant when log arguments are passed",
+				"The message given to '{0}' is built at runtime while log arguments are also passed. Every '{{...}}' in the message is read as a placeholder, so interpolated text - an exception message or a json fragment, say - can silently consume the arguments. Make the message a constant and pass the values as log arguments instead: Log.Error(\"failed {{message}}\", ex.Message).",
+				Category_Logging,
+				DiagnosticSeverity.Warning,
+				helpLinkUri: DocsPageHelper.GetCliDocsPageUrl($"{LOGGING_GUIDE_BASE_URL}#log-message-template-must-be-a-constant", Constants.CLI_CURRENT_DOCS_VERSION),
```
mhijaziB: Codex (Worth double checking)

These two help links currently point to headings that do not exist on the `CLI 7.0` troubleshooting page:
- `#log-message-template-must-be-a-constant`
- `#log-message-template-cannot-be-rendered`
Both links land at the top of the generic troubleshooting page. Can we publish the matching documentation with this change, link to an existing logging section, or omit the help links until those sections are available?

mhijaziB (commented):
The initializer call-site fix itself is going in the right direction, and the focused runtime tests pass. I would still hold the merge on the runtime fallback and analyzer-array cases above.
Right now the safe logger can still throw while formatting its fallback, and the compile-time analyzer misses explicit or forwarded object[] arguments. Those are both reproduced failures rather than theoretical concerns. The existing `SourceGen` suite passes, but it does not contain tests for the new analyzer.

mhijaziB:
@Leinnan Codex:
1. most of the original feedback has been addressed, 2 issues remain

2. [P2] Forwarded params wrappers remain invisible to the analyzer
On [LogTemplateAnalyzer.cs:80](https://github.com/beamable/BeamableProduct/blob/20db5277546588054715b2721bcca2a10d7607b3/templates/MicroserviceSourceGen/MicroserviceSourceGen/Analyzers/LogTemplateAnalyzer.cs#L76-L85):
`IsForwardedTemplate` suppresses the diagnostic on the only call this analyzer recognizes. Calls to wrappers such as `MyLog` or `MicroserviceDebug` are not analyzed because `IsBeamableLogMethod` only accepts `BeamableLogger` and `Log`. That means a caller can pass a runtime-built message with arguments through the wrapper and neither the wrapper nor its caller reports `BEAM_LOG_0001`.
Please either keep the diagnostic at this call, restrict the suppression to known-safe wrappers whose callers are analyzed elsewhere, or extend the analyzer to recognize wrapper call sites.

3. [P2] Avoid formatting the failed queued state again
On [LogUtil.cs:82](https://github.com/beamable/BeamableProduct/blob/20db5277546588054715b2721bcca2a10d7607b3/microservice/beamable.tooling.common/LogUtil.cs#L80-L95):
`SafeToString(message.state)` calls `FormattedLogValues.ToString()`, which invokes the same formatter that just failed. A throwing or side-effecting argument therefore gets formatted again during recovery. After that, the structured-state branch replaces the result with only `{OriginalFormat}`, dropping all original argument values.
Could we inspect `IReadOnlyList<KeyValuePair<string, object>>` first and construct the fallback from the original template plus safely rendered non-format entries? `SafeToString(state)` should only be used for non-structured state. A focused test with one throwing argument, one safe argument, an attached exception, and a following queued record would cover the full recovery path.

4. One non-blocking test note: the new `*Format` tests use {0} with one argument and {0} {1} with two. Those pass even if the composite-format branch is accidentally removed. A differentiating test such as `LogErrorFormat("positional {1}", new object[] { 1 })` expecting `BEAM_LOG_0002` would make the regression guard meaningful.

microservice/beamable.tooling.common/LogUtil.cs
allister-beamable: Could @mhijaziB 's point about losing the original failure be addressed by a narrower exception type in the `catch`? We know specifically which exception occurs when there is an unexpected curly-brace expression in the message, right?

allister-beamable (approved):
This was a lot more code than I expected for this issue, but perhaps that just means I underestimated the complexity of the issue. That, and most of the new code is in tests and the static analyzer: both very beneficial with no run-time cost.

Address Moe's comment about the exception-catching during formatting if possible. Other than that, good to go!

mhijaziB:
Ok to merge now, all the feedback has been addressed.

@allister-beamable I know you love details so:
I think the broad catch is intentional here. A composite `ILogger` can surface formatting failures as an `AggregateException`, and argument formatters or `ToString()` implementations may throw exceptions other than `FormatException`. Narrowing the catch would allow some logging failures to escape and potentially strand the startup queue. The recovery path now preserves the original metadata and structured values safely, while the nested catch ensures the queue still drains if the target logger itself refuses the fallback write.

mhijaziB (approved):

Issue #4806, opened 2026-08-19
Bulk local content changes can take 7–10 seconds to appear in the Unity Content window


mhijaziB:
## Describe the Bug

This is a performance and observability follow-up to #4642.

After installing the CLI build containing the filesystem-watcher overflow recovery (`Beamable.Tools 0.0.123.11`), deleting approximately 2,263 local content files eventually updated the Unity Beam Content window with the correct deleted count without a manual refresh. However, the update took approximately 7–10 seconds.

Correctness is now working in the tested case, but large local content changes are still slow and it is difficult to confirm whether the watcher overflow/full-rescan recovery path was used.

## To Reproduce

1. Open a Unity project connected to a realm with 1,000–2,000+ content entries.
2. Open the Beam Content window and allow `content ps --watch` to initialize.
3. Delete all or most local content JSON files in one operation.
4. Measure the time between the filesystem operation completing and the Beam Content window showing the correct deleted count.
5. Inspect the running `content ps --watch` command in CLI Debugger.

## Actual Behavior

- The final deleted count is correct, but the Content window takes approximately 7–10 seconds to update for roughly 2,263 deletions.
- The recovery warning was not visible during the first test: `Content filesystem notifications were lost. Rebuilding the full local content state.`
- Clearing the CLI Debugger command data before the test removes the running command from the visible history. Later logs continue attaching to its internal command descriptor, so the debugger remains visually empty.
- The warning is currently written only after the full rescan has completed, so it cannot provide immediate feedback that recovery has started.

## Expected Behavior

- Large local content operations should reconcile efficiently and should not perform an expensive partial update immediately followed by a full rebuild.
- Overflow recovery should be observable as soon as it begins.
- Clearing CLI Debugger data should not make future logs from an active command inaccessible.
- The Content window must continue to reach the authoritative on-disk state without requiring manual Refresh.

## Findings So Far

Confirmed from the current implementation:

- The watcher waits a fixed 100 ms after the first notification and then drains the channel once.
- A large filesystem operation can continue producing notifications after that drain, allowing a partial incremental batch to be processed before a later overflow sentinel.
- Each individual create/change/delete/rename currently emits a verbose JSON log.
- Unity invokes the CLI with verbose log streaming enabled, so a 2,000+ file operation can generate thousands of additional streamed log messages.
- Exact-ID filtering currently performs repeated linear membership checks for large batches.
- A full-rebuild event requires the Unity integration to replace and rebuild its content caches.
[...]

PR #4807, opened 2026-08-20
Fix Content window recovery after filesystem watcher overflow

### LINKED ISSUE #4642: Beam Content Window UI not updating properly
## Describe the Bug 

This is most obvious when deleting lots of content locally, the window doesn't fully update the `deleted` status count. For example, in a realm of 2k+ content, the UI might only show 500 `deleted` instead of showing the correct amount. This could be a general bug around UI update and not specific to content deletion, this could happen when lots of contents are getting updated.

1. Have a Realm with 1k+ content
2. Delete local content
3. Go back to Beam Content window and check status count

### Expected Behavior

Beam Content UI update properly and efficiently to show the correct statuses.

## Screenshots

Image

## Metadata

- SDK version: 5.1 (Although this happened in 5.0.1 as well)
- Engine version: Unity 6.3
- Operating System: W 11

mhijaziB:
Fixes #4642 

# Brief Description

Large local content operations can overflow the filesystem watcher's internal buffer. When this happened, some file notifications were lost and the Unity Content window displayed an incomplete deleted count until the user manually refreshed it.

This change:

- Converts watcher overflow errors into a coalesced full-rescan request for the affected PID and manifest.
- Rebuilds the local content state from disk instead of applying an incomplete incremental update.
- Emits a full-rebuild event so Unity replaces its cached content state, including when the rebuilt manifest is empty.
- Handles files disappearing while a rescan is running.
- Prevents duplicate content IDs when local files are deleted during reconciliation.
- Adds focused automated coverage for the overflow and full-rescan recovery path.

## Verification

- `ContentWatcherRecoveryTests`: 8/8 passed.
- CLI build completed with 0 errors.
- Manually tested deleting 2,000+ local content files in Unity.
- The Content window reached the correct deleted count after an overflow without requiring a manual refresh.

## Follow-up

Overflow recovery currently takes approximately 7–10 seconds in the large-content reproduction. That performance and observability work is intentionally outside the scope of this fix and is tracked separately in [#4806 — Bulk local content changes can take 7–10 seconds to appear in the Unity Content window](https://github.com/beamable/BeamableProduct/issues/4806).

# Checklist

# Notes


cli/cli/Commands/Content/ContentPsCommand.cs
```diff
@@ -294,6 +297,48 @@ void OnMessage(ContentProgressUpdateData data)
...
+
+		var reconciledContent = await fullRescan();
+		return CreateFullRebuildEvent(reconciledContent.ContentFiles, cid, pid, manifestId);
+	}
+
+	private static ContentPsCommandEvent CreateFullRebuildEvent(IEnumerable<ContentFile> contentFiles, string cid, string pid,
+		string manifestId)
+	{
+		var entries = ContentService.ContentFileToLocalContentManifestEntries(contentFiles)
+			.ToArray();
```
allister-beamable, line 320: Is this the part that is slow when there are lots and lots of content entries?

cli/cli/Services/Content/ContentService.cs
```diff
@@ -171,6 +175,69 @@ public ContentService(CliRequester requester, ConfigService config, IContentApi
...
+			changedContentFiles.TryWrite(recoveryChange);
+			return;
+		}
+
+		logger.LogError(
+			exception,
+			"Content filesystem watcher failed. PID={Pid}, ManifestId={ManifestId}",
+			pid,
+			manifestId);
+	}
```
allister-beamable, line 215: Something about this watcher error handler makes my brain itch, like it is a bridge too far in terms of moving parts. I see that we already have a filesystem watcher as part of the mechanism, but something bothers me about that.

Or, alternatively, could we make it clearer to the user that something is busted in a way that refreshing will fix, and give them a big prominent refresh button to fix it?

Third-alternatively, would it make sense to just have a heuristic where if you delete a bunch of junk the system triggers the full refresh when the volume of change exceeds a certain threshold (say 500 entries or 1000 entries)?

If these worries of mine spark thoughts for improvements, purpose of review achieved. If, on the other hand, you think this way is still the best way, go with it.

allister-beamable (approved):
So, generally I think the changes are in a good direction, with one sort of architectural concern: it feels like we are adding moving parts to an already-complex system in hopes of making it more robust.

In the design tension of simplicity vs. featureful functionality, I think we are already on the "make it shiny and reactive" side of the bridge, so we may have to reserve true simplification for the moderate to far future.

All this to say: ship it! (but look for opportunities to simplify _if possible_)

PR #4810, opened 2026-08-24
Change - Update the dev.sh and the local init to make the no web registry interactive in the local stack

felipe-beamable:
# Brief Description

> Update the dev.sh and the local init to make the no web registry interactive in the local stack

### DIFF
--- a/cli/cli/Commands/LocalStack/LocalStackInitCommand.cs
+++ b/cli/cli/Commands/LocalStack/LocalStackInitCommand.cs
@@ -22,10 +22,18 @@ public class LocalStackInitCommandArgs : CommandArgs
 	public bool updateServices;
 
 	/// <summary>
-	/// Opt OUT of writing the web-registry steps. Stored inverted so the default (a plain <c>bool</c> field,
-	/// false) means "include them" without depending on how the option binder treats an absent flag.
+	/// Answer "no" to the web-registry question without being asked. The steps are still written to the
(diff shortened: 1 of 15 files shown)


cli/cli/Commands/LocalStack/LocalStackInitCommand.cs
```diff
@@ -22,10 +22,18 @@ public class LocalStackInitCommandArgs : CommandArgs
...
+	/// <see cref="LocalStackConfig.webRegistry"/>).
 	/// </summary>
 	public bool noWebRegistry;
+
+	/// <summary>
+	/// Answer "yes" to the web-registry question without being asked. Its own field rather than an inversion
+	/// of <see cref="noWebRegistry"/> so "not passed" stays distinguishable from "passed off" — which is what
+	/// lets the interactive prompt know it still has a question to ask.
+	/// </summary>
+	public bool withWebRegistry;
```
Leinnan: TBH I do not like the idea of having data structure that could have potentially two values that contradict each other. Could you change it into one field with enum value (Unspecified/WithRegistry/WithoutRegistry) instead? It does not have to be enum exactly, just changing it in a way that there is no option to create a invalid structure.

felipe-beamable: You are right, I made a refactor, because this is part of a old requirement. Thank you.

DiasAtBeamable (approved):

PR #4811, opened 2026-08-26
Do not trigger most of the jobs on the web only change

### LINKED ISSUE #4414: Trigger Only Web SDK Jobs When PRs Modify the Web Directory
At the moment, any change made within the web folder triggers all jobs and tests (Unity, microservices, etc.) when a pull request is opened. Only Web SDK related jobs should run for changes scoped to the web directory.

We should add the path filtering to the buildPR.yml file, so that we don't kick off all jobs when changes are made outside of the main C# folders.

Leinnan:
Fixes #4414

### DIFF
--- a/.github/workflows/buildPR.yml
+++ b/.github/workflows/buildPR.yml
@@ -9,6 +9,7 @@ on:
     paths-ignore:
       - 'rfc/**'
       - 'client_installer/**'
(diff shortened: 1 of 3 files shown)


allister-beamable (approved):
Yes good. Clean and simple.

PR #4821, opened 2026-09-02
Adopt the reworked bundle catalog API in the web SDK and CLI

dpatekar:
# Brief Description

BeamableApi changed the bundle catalog contract: visibility now belongs to a bundle's name rather than to each
published checksum, the publish history became a paged release log, and dependencies carry a version range. This branch brings the two consumers in this repo up to that contract.

The web SDK is regenerated against dev, taken whole as previous regens were. Two bundle functions are renamed as a result: `beamoGetBundlesHistory` is now `beamoGetBundlesReleases`, and `beamoPutBundlesChecksumsAcl` is now `beamoPutBundlesAcl`. The regen also needed a small fix to the generator, which split path segments on `/` and `-` only and produced an invalid type name for a route ending in `.gif`.

The CLI's generated client is updated for the bundle types only, and the commands follow: `bundles history` becomes `bundles releases`, `bundles acl` takes a name instead of a checksum, and `bundles list` shows per-name summaries. The authored bundle file renames `peerDependencies` to `bundleDependencies` with a `{min, max}` range per entry.

Nothing here is shipped yet, so there is no compatibility layer. The renamed routes were verified against dev.

# Checklist


gabrielbeamable (approved):
looks good to me, just a few questions:

I see you already generated the web sdk open api files, is there a RC release for this? while we don't do a definitive release of the web sdk, we've been releasing some RCs so to use these changes in Console and portal extensions

These API changes in the web sdk might break things in the Console, specifically the bundles view page, for history etc, is there something on the mends for this? If not, that is ok, I think it shouldn't block this PR and I can take on that work

Issue #4834, opened 2026-09-08
C#MS: outbound platform requests have no timeout; a lost reply hangs the caller forever (MicroserviceRequester TODO)


mhijaziB:
## Problem

Outbound platform calls from a C# microservice have no deadline. `MicroserviceRequester.BeamableRequest` registers a promise in `SocketRequesterContext._pendingMessages`, writes the frame, and returns. The promise completes only when a reply frame with the same id comes back. If the platform never answers, or the socket drops mid-flight, the promise is orphaned forever. Nothing sweeps the pending dictionary on disconnect either; the only "noconnection" failure path is a reply with status 0, which the platform has to send.

The gap is acknowledged in the code:

```csharp
// microservice/beamable.tooling.common/Microservice/MicroserviceRequester.cs
// TODO how do we handle Timeout errors?
// TODO what does concurrency look like?
```

The existing knobs do not cover this. `REQUEST_TIMEOUT_SECONDS` is a cooperative cancellation token for inbound requests. `WaitForAuthorization` caps the wait for the auth daemon before a send, not the wait for a reply. The `X-KS-TIMEOUT` header raises the platform's budget and does nothing locally.

## Why it matters

A dropped or lost reply becomes an indefinite hang with no log line and no exception. Callers that wrap the call in their own `Task.Delay` race see a `TaskCanceledException` at their own deadline and blame latency; callers that do not wrap it wait forever. We have seen this shape in production (#4833). The timeout does not make the request succeed, it makes the failure visible and bounded.

## Proposed change

1. **Deadline per pending request.** Stamp each listener with a creation time. One periodic sweeper per socket, not one timer per request, walks `_pendingMessages` every second and fails anything past its deadline with a dedicated `RequesterException` subtype carrying method, path, request id, and elapsed time. Existing `catch (RequesterException)` blocks pick it up unchanged.
2. **Default above the platform budget.** The relay gives a call 10s and answers with a real 504 when a downstream service is slow. The local deadline is a backstop for the no-reply case, so default to 30s. When the caller sets the timeout header, follow it plus a margin. Expose an environment variable for the default; `0` disables it for anyone relying on very long calls.
3. **Drain on disconnect.** In `HandleCloseConnection`, fail every pending promise with the existing "noconnection" error. A reconnected session has no memory of the old ids, so those requests can never be answered.
[...]

PR #4835, opened 2026-09-08
Fix - Omit empty inventory scope instead of sending `?scope=` (#4833)

### LINKED ISSUE #4833: C#MS Services.Inventory.GetCurrent() with empty scope hangs forever on runtime 7.1.0+ (sends `?scope=`, gateway never replies)
## What happens

On Beamable.Microservice.Runtime 7.1.0 and later, `Services.Inventory.GetCurrent()` with no scope never returns. The promise hangs forever; there is no error and no log line. Players see a 504 once the platform relay gives up.

## Why

PR #4535 (shipped in 7.1.0) rewrote `MicroserviceInventoryApi.GetCurrent` to call the autogenerated `IInventoryApi.ObjectGet(userId, scope)` with the raw `string scope`. The implicit `Optional<T>(T)` conversion sets `HasValue = true` for an empty string, so the request goes out as `object/inventory/{id}/?scope=`. The gateway drops any request with an empty query value without replying (beamable/BeamableBackend#931). `MicroserviceRequester` has no outbound timeout (`// TODO how do we handle Timeout errors?`), so nothing ever fails the promise.

7.0.1 built the URL by hand and omitted `scope` when empty, which is why it still works.

## Evidence

- A client: every CoreService inventory health check hung at 100% on 2026-09-01 (7.2.2) and 2026-09-03 (7.2.3) for the life of the instances. Both episodes ended only when they redeployed on 7.0.1.
- Reproduced 2026-09-07 with a scratch service on 7.2.3 against prod: `ObjectGet(id, OptionalString.None)` replies in 239ms, `ObjectGet(id, "currency")` in 226ms, `ObjectGet(id, "")` gets no reply after 15s, and the next call on the same socket replies in 228ms.

Helpers that pass a real scope (`GetItems<T>()`, `GetCurrency`, `GetCurrencies`) are not affected. Only a bare `GetCurrent()` or `GetCurrent("")` hits it.

## Fix

1. `MicroserviceInventoryApi.GetCurrent`: pass `OptionalString.None` when `scope` is null or empty.
2. Audited 2026-09-08: of the 138 autogenerated methods that build a query string, the C#MS runtime wrappers call four (`InventoryApi.ObjectGet`, `AnnouncementsApi.ObjectGet`, `EventPlayersApi.ObjectGet`, `BeamoApi.GetServicesLogsContext`). Only the inventory one passes a raw string into an `Optional<string>` query parameter, so this is the single affected site in the SDK. The trap itself is generic though: any C#MS code calling an autogenerated API directly with `""` for an `Optional<string>` query parameter hits the same drop. Worth a guard in the generated code (skip empty optional query values) or in `Optional<T>`'s implicit conversion.
3. Requester timeout hardening, tracked separately in #4834 so it ships on its own (see below).

## Workaround

Pass a non-empty scope (`GetCurrent("items")`, `GetCurrent("currency")`), or stay on 7.0.1.

## Why the hang was indefinite (hardening, #4834)

`MicroserviceRequester.BeamableRequest` registers a promise in `_pendingMessages`, writes the frame, and waits for a reply with the same id. There is no deadline and nothing drains the dictionary on disconnect; the code marks the gap with `// TODO how do we handle Timeout errors?`. `REQUEST_TIMEOUT_SECONDS` only covers inbound requests and `WaitForAuthorization` only covers the auth wait before a send.

#4834 proposes: a per-request deadline enforced by one sweeper per socket, a 30s default that sits above the platform's 10s relay budget and follows the `X-KS-TIMEOUT` header when set, configurable via environment variable with `0` to disable; drain pending promises on disconnect with the existing "noconnection" error; downgrade the late-reply "no listener" log to debug for expired ids; error status on the OpenTelemetry activity plus a counter.

It turns a silent hang into a bounded, visible failure. It does not make the request succeed; that is the platform fix in beamable/BeamableBackend#931. Keep it out of the scope patch so the fix that unblocks upgrades carries one behavioural change.

mhijaziB:
Fixes #4833.

`Services.Inventory.GetCurrent()` and `GetCurrent("")` now omit the empty scope query parameter. This avoids the request shape associated with inventory calls remaining pending on runtimes 7.1.0 through 7.2.3. `GetCurrent(null)` also treats the scope as omitted; previously it threw before sending a request. Non-empty scopes keep their existing behavior.

## Cause and evidence

PR #4535 changed `MicroserviceInventoryApi.GetCurrent` to pass its raw scope string into the autogenerated `IInventoryApi.ObjectGet`. The implicit conversion to `Optional<string>` marks an empty string as present, so the generator sends `object/inventory/{id}/?scope=`. A null string also becomes a present optional, but then `scope.Value.ToString()` throws before the requester is called.

The requester registers a response listener with no outbound response deadline. If no matching response arrives, the promise remains pending. This code path is verified. It does not establish where the platform loses the request or response.

The production probe recorded in beamable/BeamableBackend#931 reported no reply after 15 seconds for `?scope=`, while omitted scope and `scope=currency` returned in roughly 230 ms on the same socket. We did not rerun that production probe during this review. The precise platform failure location remains unproven; the gateway query parser is the hypothesis tracked in that issue.

Before 7.1.0, `GetCurrent` used `BeamableGetApiResource.RequestData` and its URL builder in `ISupportsGet.cs`, which omitted null or empty scope. We restore that omission, not the exact previous URL bytes: the generated route still has a trailing slash.

## Scope and remaining work

- The empty-reference-array overload of `GetItems<T>(...)` joins its references into an empty scope and reaches the corrected `GetCurrent` method, so it also benefits from this fix.
- `GetCurrency("")` and `GetCurrencies(new[] { "" })` still construct a present empty scope independently of `GetCurrent`. We should validate each supplied currency ID before sending and add tests that assert no outbound request for invalid IDs. The existing null/empty currency-array behavior, which requests all currencies, should remain unchanged.
- Direct generated calls such as `IInventoryApi.ObjectGet(id, "")` remain exposed. Callers can use `OptionalString.None` to omit scope. A broader SDK guard needs an endpoint-specific contract; changing all optional strings could discard meaningful empty values.
- Gateway request/response handling is tracked in beamable/BeamableBackend#931. Outbound timeout and pending-listener cleanup are tracked separately in #4834. Neither change is included here.
- Unity's normal inventory `GetCurrent` path uses a POST helper with scopes in the body. The CLI references the microservice runtime project; this patch changes the runtime inventory wrapper, not CLI command logic.

## Verification

- Five `InventoryScopeTests` cases cover default, empty, and null scope; an empty item-reference array; and a non-empty currency scope. Each checks the outgoing route and the client reply.
- `TestSetupShutdownTests` proves that the shared shutdown helper waits for the service. This regression test failed before the harness fix and passed afterward.
- The inventory fixture waits for an explicit client reply and awaits shutdown in `finally`, replacing the fixed 50 ms sleeps.
- Focused local run: 6 passed, 0 failed.
- Full local `microserviceTests` suite: 280 passed, 0 failed, 0 skipped.

### DIFF
--- a/microservice/microservice/Api/Inventory/MicroserviceInventoryApi.cs
+++ b/microservice/microservice/Api/Inventory/MicroserviceInventoryApi.cs
@@ -19,9 +19,8 @@ namespace Beamable.Server.Api.Inventory
 
 		public override Promise<InventoryView> GetCurrent(string scope = "")
 		{
-			// An empty scope means "the whole inventory". It must be omitted from the request rather than sent as
-			// an empty query value (`?scope=`): the gateway never replies to that request, and the promise would hang
-			// forever. This matches the pre-7.1.0 behaviour, which only appended the scope when it was non-empty.
+			// Null or empty scope means "the whole inventory". Omit the query parameter, as the pre-7.1.0
+			// BeamableGetApiResource did. Sending `?scope=` can leave the websocket request pending (#4833).
 			Optional<string> scopeArg = string.IsNullOrEmpty(scope)
 				? OptionalString.None
 				: new OptionalString(scope);
--- a/microservice/microservice/Api/Inventory/MicroserviceInventoryApi.cs
+++ b/microservice/microservice/Api/Inventory/MicroserviceInventoryApi.cs
@@ -19,7 +19,14 @@ namespace Beamable.Server.Api.Inventory
 
 		public override Promise<InventoryView> GetCurrent(string scope = "")
 		{
-			return InventoryApi.ObjectGet(UserContext.UserId,scope).Map(InventoryViewToAutoGenInventoryView);
+			// An empty scope means "the whole inventory". It must be omitted from the request rather than sent as
+			// an empty query value (`?scope=`): the gateway never replies to that request, and the promise would hang
+			// forever. This matches the pre-7.1.0 behaviour, which only appended the scope when it was non-empty.
+			Optional<string> scopeArg = string.IsNullOrEmpty(scope)
+				? OptionalString.None
+				: new OptionalString(scope);
+
+			return InventoryApi.ObjectGet(UserContext.UserId, scopeArg).Map(InventoryViewToAutoGenInventoryView);
 		}
 
 		public async Promise SendCurrency(Dictionary<string, long> currencies, long recipientPlayer,
--- a/microservice/microservice/CHANGELOG.md
+++ b/microservice/microservice/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- `Services.Inventory.GetCurrent()` with an empty or null scope no longer sends an empty `?scope=` query value, which the gateway never answered and left the call hanging forever on runtime 7.1.0 and later.
+- `Services.Inventory.GetCurrent()` and `GetCurrent("")` now omit the empty scope query parameter, which could leave inventory requests pending on runtimes 7.1.0 through 7.2.3. `GetCurrent(null)` also treats the scope as omitted instead of throwing before sending the request.
 
 ## [7.2.3] - 2026-08-26
 
--- a/microservice/microservice/CHANGELOG.md
+++ b/microservice/microservice/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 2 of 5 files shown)


allister-beamable (approved):
MicroserviceInventoryApi change looks good. What's with the shutdown test? Was it an incidental touch? Fine to include it, in my opinion, but it seems extraneous to the fix itself.

mhijaziB:
@allister-beamable Yes, it was an incidental test-harness fix we caught while adding the inventory regression tests. `TestSetup.OnShutdown` discarded the service’s shutdown task, so awaiting it didn’t actually guarantee cleanup had finished before the next test. We fixed that and added the shutdown test to prevent it regressing. We also replaced the inventory tests’ 50 ms sleeps with an explicit wait for the reply. You’re right that it’s separate from the inventory fix itself.

PR #4840, opened 2026-09-14
Derive content reference checksums locally instead of trusting publishers

### LINKED ISSUE #4838: Content Manager's "modified" flag should be resilient to JSON field ordering inconsistencies
## Describe the Bug 

One Game Maker reports that after fetching content, the Content Manager tab shows entries as modified even though their field values are identical. When the developer inspected the files on disk and compared them to the content shown in the Beamable Portal, they found that the JSON differed only by field order: that is, the field keys and values were indeed identical, but the order in which the key/value pairs appeared in the JSON object was different.

One contextual wrinkle: the developer uses an in-house tool that publishes content independently from the Beamable Unity SDK.

1. Go to the Beamable Content Manager tab within Unity
2. Press "Fetch"
3. Wait for the download to finish
4. Observe the badges in Content Manager and log messages in Console

### Expected Behavior

 - Zero "modified" badges; content appearing clean and unmodified.

### Actual Behavior

 - Persistent "modified" badges even on value-identical content. JSON on disk differs from API-supplied JSON in field order.

## Metadata

- SDK version: Beamable Unity SDK 6.0.1 / CLI 7.2.2
- Engine version: any Unity
- Operating System: macOS and Windows

## Additional Context

We fixed similar issues in the past by enforcing a sort order on the JSON.

allister-beamable:
Closes #4838.

## The problem

`ContentFile.GetStatus()` decided whether content was modified by comparing two values that are not comparable:

- `PropertiesChecksum`, computed locally by `CalculateChecksum` over our own canonical serialization, and
- `ReferenceContent.checksum`, a string lifted verbatim out of the remote manifest.

The second one is whatever the publisher put there. Nothing recomputes or validates it. So the real contract was never "sort your keys" — it was "byte-for-byte reproduce `System.Text.Json` under one undocumented options object, then SHA1 it." A Game Maker publishing through their own tooling cannot honor that except by accident, and when they do not, every item shows a permanent "modified" badge that survives a Fetch.

Field order is only one of the ways to miss. Others include string escaping (`JavaScriptEncoder.Default` renders `Café & Co` as `Caf\u00E9 \u0026 Co`), raw number text (`1.0` and `1` both round-trip as written), and keys differing only in case.

## The change

**Stop trusting the publisher's checksum.** Content files gain a `reference` object holding what the remote payload hashed to under *our* canonicalization, together with the platform content version identifying which payload that was:

```json
"reference": { "checksum": "5c1f…", "version": "9ab3…" }
```

`GetStatus()` compares against that checksum when the version matches the manifest entry being compared against, and falls back to the old comparison otherwise. Both sides of the comparison then go through the same serializer, so the result no longer depends on how the publisher formatted its JSON.

The version pairing is load-bearing rather than decorative. The platform derives a content version from the payload itself, so it identifies exactly which remote bytes our checksum describes. Without it, a file synced from an older manifest would compare its stale reference against a newer manifest entry, match, and report a genuinely changed remote as up to date — trading a false positive for a false negative, which is much worse. There is a test for exactly that.

The two halves are meaningless apart, so they are never apart. `LocalContentReference` can only be constructed whole, `TryRead` yields nothing unless both halves are present, and the JSON nests them under one key, so a half-populated reference is unrepresentable rather than guarded against. Absence is a null, not a pair of empty strings.

The reference is written when we download content, when content we publish becomes the new remote (using the version the platform returns on the save response), and when a file that genuinely matches the target has its manifest reference moved forward. Files without one fall back to the old manifest-checksum comparison, so nothing changes for existing workspaces until their next sync.

**Pin the serializer options.** `GetContentFileSerializationOptions` now names `Encoder = JavaScriptEncoder.Default` explicitly. This is what `System.Text.Json` was already doing implicitly, so **no existing checksum changes** — it just can no longer drift with a .NET upgrade, and the comment says why that matters.

**Make the key sort a total order.** `OrderBy` is stable over input ordering, so two keys differing only in case previously kept whatever order they arrived in, which is exactly the instability the converter exists to remove. Both sorted converters get an `Ordinal` tiebreak. In `SortedJsonElementConverter` the tie was a correctness problem, because its output is hashed. In `SortedSnapshotConverter` it was not -- snapshot bytes are not hashed -- but an incidental reordering there turns a one-line diff into a two-line one, which defeats the readability the sort was introduced for. **No existing checksum changes**, since only case-colliding keys reorder.

## Deliberately not done

- **Numbers are not normalized.** Raw text replay is kept. Normalizing risks precision loss on doubles and erases the float-versus-int distinction Unity schemas rely on, and the local-reference design removes most of the exposure anyway, since the local file is written from the downloaded payload through the same serializer. Residual: a Unity edit that rewrites `1` as `1.0` can still read as modified. Covered by a test that documents the behavior as intentional.
- **The encoder was not switched to relaxed escaping.** It would make content files far more readable in diffs, but it changes every checksum for content containing non-ASCII or `& < > '`, which means a forced re-sync and skew against teammates on older CLI versions. Worth doing deliberately with a migration, not as a side effect of this fix.
- **Snapshot restore does not repopulate the reference.** Snapshots record the local checksum at snapshot time, not the reference, so it cannot be recovered from one. Restored files fall back to the legacy comparison until their next sync. Tracked in #4852, which also covers repairing snapshots taken before this change.

## Server-side half, not addressed here

Content versions are derived from a hash of unsorted JSON on the platform side as well, so a value-identical republish with different key order mints a new content version, a new stored object, and a new manifest. That means an unstable publisher still causes churn and redundant client downloads even with this fix in place. It is filed separately on the backend tracker; it needs a migration plan because existing stored object keys derive from the current hash, and it is not something the CLI can fix on its own.

## Tests

`ContentChecksumTest` covers the canonical form as a wire format: top-level and nested field order are ignored, case-only key collisions order deterministically, array order is still significant, different values still differ, the pinned encoder still escapes exactly as before, and raw number text is still preserved.

`ContentStatusTest` covers what `GetStatus()` trusts: an unreproducible publisher checksum no longer marks value-identical content modified, a remote that moved on since we recorded our reference still reads as modified, genuinely different local properties still read as modified, and a file carrying no reference falls back to the historical comparison. It also pins the on-disk shape — the reference nested under one key, and omitted entirely when absent — and checks that a half-written reference is refused at parse time.

The tiebreak added to `SortedSnapshotConverter` is not separately covered. No existing test pins snapshot key order, and its output is not hashed, so a regression there would show up as diff churn rather than as wrong status. Say the word if you would rather it were pinned.

## Notes for review

- Measured cost of the added hash: roughly 36 µs per item at ~1.9 KB of properties on an M5 Pro, of which SHA1 itself is under 5% — the sorted serialization dominates. It runs once per downloaded item over an already-parsed payload, inside the existing `Task.WhenAll`, against a sync that issues one CDN request per item. If it ever needs to be cheaper, `CalculateChecksum` currently serializes to a `string` and then calls `Encoding.UTF8.GetBytes`; serializing into a pooled `ArrayBufferWriter<byte>` and hashing the span measured about 37% faster at that size.
- No runtime or on-device impact. The whole checksum apparatus in `ContentObject.cs` sits inside `#if UNITY_EDITOR`, and the CLI is not shipped to devices.


cli/cli/Services/Content/ContentService.cs
```diff
@@ -1382,6 +1397,11 @@ public async Task<ContentSyncReport> SyncLocalContent(ClientManifestJsonResponse
 			contentFile.Properties = j.GetProperty("properties");
 			contentFile.Tags = JsonSerializer.SerializeToElement(c.ReferenceContent.tags);
 			contentFile.FetchedFromManifestUid = targetManifestUid;
+
+			// Hash the downloaded payload ourselves rather than trusting the publisher's manifest checksum. The
+			// file we are about to write is serialized from this same element, so the two agree by construction.
+			contentFile.PropertiesChecksum = CalculateChecksum(in contentFile);
+			contentFile.Reference = new LocalContentReference(contentFile.PropertiesChecksum, c.ReferenceContent.version);
```
mhijaziB: Could we also apply this reference update in `ContentResolveConflictCommand` when `--use realm` downloads content? That path replaces the properties and manifest UID but retains the old reference checksum/version. With a publisher checksum that differs from ours, the downloaded content still reports Modified after resolving with the realm version. That path should calculate and save the new reference too.

cli/cli/Services/Content/ContentService.cs
```diff
@@ -2188,12 +2226,56 @@ public struct LocalContentFiles
...
+	}
+
+	/// <summary>
+	/// Reads a reference, yielding one only when both halves are present.
+	/// </summary>
+	public static bool TryRead(in JsonElement json, out LocalContentReference reference)
+	{
+		reference = default;
+
+		if (!json.TryGetProperty(JSON_NAME_CHECKSUM, out var checksumElement)) return false;
```
mhijaziB: Could we check `json.ValueKind == JsonValueKind.Object` before this call and validate the member types before calling `GetString()`? An explicit `"reference": null` currently throws `InvalidOperationException` instead of falling back to the legacy comparison. The exception propagates through `GetAllContentFiles`, so one such file fails the content scan. Please add coverage for null and incorrectly typed reference values.

mhijaziB (commented):
The approach makes sense. Before merging, we should refresh the reference when resolving conflicts using the realm version and handle null or malformed reference metadata without failing the content scan. Details in the two inline comments.

mhijaziB:
@allister-beamable  the null/malformed metadata handling is fixed. The remaining gap is `ContentResolveConflictCommand: --use` realm still downloads the properties without refreshing the baseline, so identical content can remain marked Modified.

cli/cli/Services/Content/ContentService.cs
allister-beamable: I am moving towards the reference being a single string, so this will be moot. But also, how would there ever be `"reference": null` in there? It would take manual chopping of the JSON to even produce that, and anyone hitting a big ol' `InvalidOperationException` immediately after writing `null` into a file might (one would hope) think, "oh, I bet the null I just wrote caused that".

PR #4842, opened 2026-09-15
Bump mongo driver deps

### LINKED ISSUE #4841: Bump MongoDB.Driver to 3.11.2
Old version is using Snappier in version that is vulnerable, change requires updating `SharpCompress` library as well.

Leinnan:
# Ticket

Resolves #4841


cli/cli/Services/ServiceUploadUtil.cs
```diff
@@ -597,7 +597,7 @@ static async Task<byte[]> GetBytesFromStream(Stream stream)
 		return mem.ToArray();
 	}
 	
-	static async Task<(bool, byte[])> TryGetBytesForEntry(TarArchive archive, string entryName)
+	static async Task<(bool, byte[])> TryGetBytesForEntry(IArchive archive, string entryName)
```
allister-beamable, line 600: Are these changes in `ServiceUploadUtil.cs` directly related to the MongoDB driver version upgrade? If not, I would rather they be in a separate style-cleanup commit.

allister-beamable (approved):
Is SharpCompress a hard dependency of MongoDB.Driver? If so, ship this PR as-is. If not, I think there is enough value in the rollback-ability of "one PR per dependency" to keep them separate.

PR #4847, opened 2026-09-16
XUnit dependency update

Leinnan:


allister-beamable (approved):
Looks good, but: description, please! ☀️

PR #4850, opened 2026-09-16
Update OpenTelemetry dependencies to 1.18

Leinnan:


cli/beamable.microservice.otel.exporter/MicroserviceOtelActivityExporter.cs
```diff
@@ -20,11 +20,12 @@ public MicroserviceOtelActivityExporter(MicroserviceOtelExporterOptions options)
...
 			endpoint = $"{endpoint}/v1/traces";
 		}
-		_otlpOptions = new OtlpExporterOptions
+		_otlpOptions = new OtlpExporterOptions
 		{
 			Endpoint = new Uri(endpoint),
-			Protocol = options.Protocol,
-		};
+			Protocol = options.Protocol,
+			MaxRequestSizeBytes = 128 * 1024 * 1024,
```
allister-beamable, line 27: Where does this 128 MiB amount come from? Was that already the default in OpenTelemetry 1.12.0? Could you update the PR description to mention why/how you chose 128MiB, even if it is just "arbitrary safe value"?

allister-beamable (approved):
Looks good. Please write a body for the PR so that when we look back and ask, "why this particular way?" the description answers it 😸 

Nothing quite like dependency version upgrades to remind us quite how much repetition and boilerplate occurs in our code base sometimes 😉

After writing the PR body, this one is good to go. Ship it!

cli/beamable.microservice.otel.exporter/MicroserviceOtelActivityExporter.cs
Leinnan: That was the default value in the previous version, here I just preserved that behavior.

PR #4853, opened 2026-09-18
Fix content deletion races and surface nested validation errors (#4845)

### LINKED ISSUE #4845: Content Window: first deletion is undone after a clean publish or full sync
## Problem

After a clean publish or full sync, deleting an item in the Unity Content Window does not stick on the first attempt. The item shows as Deleted for about one second, then returns to its normal state. Deleting it a second time works.

The same pattern repeats for another item immediately afterward. This is per item, not only the first delete after the operation.

We found this while testing the content, sync, publish, and Issues flows for #4839. The status and badge updates otherwise behaved as expected. This bug needs separate investigation and a fix.

## Reproduction

1. Publish pending content successfully, or perform a full sync, leaving content clean.
2. Delete an existing published item through the Content Window and confirm the deletion.
3. Observe the item briefly show as Deleted, then return to its normal state.
4. Delete the same item again. It now remains Deleted.
5. Repeat with another item. Its first deletion also reverts; the second sticks.

**Expected:** One confirmed deletion leaves the item marked Deleted until we explicitly revert/sync it or publish the deletion.

**Actual:** The first deletion is silently undone. We have to repeat the action, which makes the content state and deletion workflow unreliable.

Environment: Unity 6000.3.22f1, Windows, CLI-backed Content Window.

## Investigation so far

A delayed autosave recreating the deleted JSON file is the strongest candidate. We do not have proof yet that this is the callback sequence in the reported Unity session.

The current source permits this sequence:

- `CliContentService.DeleteContent()` marks an existing item Deleted and calls `File.Delete()`. It does not invalidate the old content object's save callback.
- `ContentObject.DelayedValidate()` invokes `OnEditorChanged` after approximately 0.3 seconds. `LoadContentObject()` connects that callback to `SaveContent(contentObject)`.
- `SaveContent(ContentObject)` suppresses identical repeated saves, but does not reject a deleted item or an obsolete object. The CLI save path can recreate the missing file.
- Publish/sync completion reloads content and clears `_lastSavedPropertiesCache`. A first stale save can therefore run; a second identical save for that ID is suppressed. This fits the repeated first-delete/second-delete pattern.

A controlled test using the current `SaveContent` and `DeleteContent` method bodies, with stubbed Unity/CLI dependencies, reproduced that pattern for two separate IDs when an old save callback was delivered after deletion. This confirms the mechanism, not the live Unity trigger. No production source or live content was changed during the investigation.

Source entry points:

- `client/Packages/com.beamable/Editor/ContentService/CliContentService.cs`: `SaveContent`, `DeleteContent`, `ClearCaches`, `CacheScriptableContent`, `LoadContentObject`.
- `client/Packages/com.beamable/Common/Runtime/Content/ContentObject.cs`: `OnValidate`, `DelayedValidate`, `CheckForNonDetectedChanges`. Shared source also lives under `cli/beamable.common/Runtime/Content/ContentObject.cs`.
- `client/Packages/com.beamable/Editor/Modules/Content/UI/ContentObjectEditor.cs`: `OnInspectorGUI`.
- `cli/cli/Services/Content/ContentService.cs`: `BulkSaveLocalContent`, `SaveContentFile`.

## Next checks

Capture one failed first deletion with the content ID, object instance, file existence, and call order around Delete, the delayed callback, CLI save, and watcher update. Confirm which operation recreates the file before choosing the fix.

Also compare deleting immediately after selection with selecting the item and waiting five seconds first. This distinguishes a selection-triggered pending save from activity triggered by deletion. That check has not been completed yet.

If autosave is confirmed, we should invalidate stale callbacks and ensure an already-running save cannot undo deletion. Preserve legitimate creation, duplication, and subsequent edits. Normal watcher deletion handling appears to preserve Deleted; a separate stale-snapshot write race in the watcher remains a secondary possibility.

## Acceptance criteria

- [ ] After a clean publish, the first confirmed deletion persists for each tested item.
- [ ] After a full sync, the first confirmed deletion persists for each tested item.
- [ ] A delayed callback or in-flight save from before deletion cannot recreate the item.
- [ ] The row, footer counts, Inspector, and Publish preview agree on the deleted state without another click or manual refresh.
- [ ] Creating, duplicating, editing, explicitly reverting/syncing, and publishing content still behave correctly.
- [ ] A regression test covers the confirmed failing event order, followed by verification in Unity.

mhijaziB:
Fixes #4845.

Content deletion now cancels pending autosaves, waits for active property/tag writes, and blocks new writes during deletion. We also prevent stale save continuations from overwriting newer tags and restore editing if file deletion fails.

Missing references now show an Inspector issue badge and errors above affected lists, even when collapsed. Inline errors remain, with soft salmon text and the original bright red bars. List validation avoids repeated parent lookups.

The shared `ContentObject.cs` change lives in CLI Common; its generated Unity copy stays in sync. Includes changelog entries and regression tests.

**Verification**
- Unity 6000.3.22f1 compiled successfully.
- 14 NUnit cases passed through direct invocation in the live Editor, plus four controlled checks covering stale tags and locked-file recovery. The standard runner hit an unsaved-scene prompt; the full suite was not rerun locally.
- Moe verified create, modify, delete, sync, and publish during development, including 20+ newly created/deleted items and resync.


cli/beamable.common/Runtime/Content/ContentObject.cs
```diff
@@ -440,17 +440,34 @@ private IEnumerator DelayedValidate()
 		     yield return null;
 	      }
 	      
-	      OnEditorChanged?.Invoke();
-	      _validateCoroutine = null;
+			_validateCoroutine = null;
```
allister-beamable: Looks like indentation changed? Not a huge deal, but I would like to try and keep the repo clean with regards to syntax style. At some point I thought we had some linter checks in CI, but maybe they wouldn't catch this. @Leinnan do you know whether we can just have a linter coerce files into style conformance when pushed to a PR's branch?

client/Packages/com.beamable/Editor/ContentService/CliContentService.cs
```diff
@@ -339,36 +431,40 @@ public async Task SyncContents(bool syncModified = true,
 
 		public void SetContentTags(string contentId, string[] tags)
 		{
+			SetContentTagsAsync(contentId, tags).Error(Debug.LogException);
```
allister-beamable, line 434: Is this meant to not `await` the `Promise`-returning call? That is, the intention is to fire-and-forget? The old way just did the whole set-tags business synchronously, right?

client/Packages/com.beamable/Editor/ContentService/CliContentService.cs
```diff
@@ -339,36 +431,40 @@ public async Task SyncContents(bool syncModified = true,
...
+		}
+
+		/// <summary>
+		/// Awaits the tag command, including direct tag edits that do not save properties first.
+		/// </summary>
+		private Promise SetContentTagsAsync(string contentId, string[] tags)
```
allister-beamable, line 440: Why `private Promise` and not `private async Promise`? Seems like misleading nomenclature for "SetContenTagsAsync" to not be marked `async` in C# terms.

client/Packages/com.beamable/Editor/ContentService/CliContentService.cs
```diff
@@ -528,27 +624,82 @@ public string RenameContent(string contentId, string newName)
...
+		}
+		
 		private bool ValidateNewContentName(string newName)
 		{
 			return Path.GetInvalidFileNameChars().All(invalidFileNameChar => !newName.Contains(invalidFileNameChar));
 		}
 
-		public void DeleteContent(string contentId)
+		/// <summary>
+		/// Retires the object immediately, waits for all its submitted writes, then removes the file.
```
allister-beamable: Could we find a synonym for "Retires" that looks less like "Retries"? Or is "retire" a highly technical term in this context?

client/Packages/com.beamable/Editor/Modules/Content/UI/ContentObjectEditor.cs
```diff
@@ -79,26 +85,27 @@ protected override void OnHeaderGUI()
 					return;
 				}
 				
-				if (contentObject.ContentStatus is not ContentStatus.UpToDate)
-				{
-					var statusIconSize = 15f;
+				const float statusIconSize = 15f;
```
allister-beamable, line 88: Style nit: we mostly use `var` for declarations rather than explicit types. Since `15f` is already clearly a `float` it should not be necessary to declare `float`. Or does `const` not have implicit typing the way `var` does?

client/Packages/com.beamable/Editor/UI/BeamGUI/BeamGUI_Utils.cs
```diff
@@ -5,6 +5,8 @@ namespace Beamable.Editor.Util
 {
 	public partial class BeamGUI
 	{
+		// Soft salmon (#F09B88) keeps content errors consistent across fields, headers, and windows.
+		internal static readonly Color ContentErrorColor = new Color32(240, 155, 136, 255);
```
allister-beamable, line 9: I thought the feedback was to use the same light gray as all other Unity text, and let the red error-bar be the sole color indicator of an erroneous condition.

allister-beamable (changes requested):
Some questions about async and Promises, and some style nits.

client/Packages/com.beamable/Editor/ContentService/CliContentService.cs
allister-beamable: Reviewing a bit deeper, the original was not strictly synchronous: it did fire-and-forget by dropping `Promise`s on the floor. This one with its `.Error()` handler is a strict improvement. Objection retracted.

client/Packages/com.beamable/Editor/ContentService/CliContentService.cs
allister-beamable: Retracted for the same "it is actually asynchronous by virtue of returning `Promise` and the `async` keyword would actually make it worse" logic as the other async objection.

client/Packages/com.beamable/Editor/Modules/Content/UI/ContentObjectEditor.cs
allister-beamable: Style rage: our prevailing convention is kinda-sorta-use-`var`-everywhere-except-kinda-not. At any rate, my assumption that `const` can be implicitly typed is flat wrong: this is correct as it stands.

client/Packages/com.beamable/Editor/UI/BeamGUI/BeamGUI_Utils.cs
allister-beamable: Reviewing the thread, @PedroRauizBeamable 's statement could easily (and probably better than my interpretation) be interpreted as "salmon is fine for the text, but bring the bar back to pure red". Retracting this objection.

allister-beamable (approved):
Gonna YOLO this into `main` now; if we need to make further adjustments on Monday, we can make new PRs.

